PR #72, opened 2015-01-02
Several ergonomics fixes for completion:

aemoncannon:
* Swallow exceptions in the candidate retriever (missing connection won't complain)
* Replace '.' with ' ' if an operator is chosen from candidates.
* Replace 'value_=' with 'value =' for setter completions, and use infix expansion for argument.
* Show type in completion menu for nullary methods.
* Expand '()' for java methods with zero params
* Manually invoke yasnippet expansion when the identifier is already complete (resolves ensime/ensime-server/#753)
* Using emacs function 'read-char-choice' to prompt for choice of paren or block expansion for certain higher order functions.
* Add tests for parameter expansion.

Questions: 
* How usable is the block prompt?

### DIFF
--- a/ensime-auto-complete.el
+++ b/ensime-auto-complete.el
@@ -21,6 +21,7 @@
 
 (require 'auto-complete)
 (require 'yasnippet)
+(require 'scala-mode2-syntax)
 
 (defcustom ensime-completion-style 'company
   "Should be either 'company or 'auto-complete."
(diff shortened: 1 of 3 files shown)


rorygraves:
@aemoncannon The CI build fails on your latest commit - http://fieldmarkeng.co.uk/job/Ensime-Emacs-CI-Build-PR-Builder/4/console

aemoncannon:
Squashed.

PR #73, opened 2015-01-10
Don't complete in comments. See ensime/ensime-server#772

aemoncannon:
It appears that recovering the comment regions in the compiler would require some work, so use a client-side hack to avoid ever making the request.


fommil:
:smile:

PR #78, opened 2015-01-19
Remove extra dash in ensime--parent-dir

actionshrimp:
Remove a couple of calls to `ensime--parent-dir`, which are only hit as a last resort when various parts of the java environment are wrong. Update to `ensime-parent-dir` as defined in `ensime-utils.el` instead.

### DIFF
--- a/ensime-util.el
+++ b/ensime-util.el
@@ -79,10 +79,6 @@
 
 ;; File/path functions
 
-(defun ensime-parent-dir (dir)
-  (unless (equal "/" dir)
-    (file-name-directory (directory-file-name dir))))
-
 (defun ensime-source-file-p (&optional filename)
   "Return t if the given filename (or the currently visited file if no
 argument is supplied) is a .scala or .java file."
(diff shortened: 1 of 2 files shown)


aemoncannon:
Lgtm

rorygraves:
ok to test

rorygraves:
FYI - "ok to test" is a command for our jenkins instance (The ensime-bot) - need to update the first message because I always have to go and lookup the right words to get it to trigger.

fommil:
@actionshrimp good spot. But can I recommend that the correct fix be to add an extra dash in the function name itself? This is good API convention to let others know not to use the function (like lisp version of "package private")

aemoncannon:
I thought -- meant module (file) private?

aemoncannon:
@fommil I changed from -- to - after a previous exchange.

fommil:
@aemoncannon it does, looks like it is in the wrong file

aemoncannon:
@fommil Gotcha. So you're saying move to ensime-vars and rename with '--'?
sgtm

actionshrimp:
I've moved ensime-parent-dir from `utils` into the `vars` module, and renamed it to the `--` version as suggested in d79fa00

fommil:
great, thanks for the contribution!

PR #80, opened 2015-01-21
Use 0.8.11 protocol changes

edani:
- completions without saving the buffer
- format-source without saving

Also unrelated changes:
- fix name of ensime inferior buffer
- do semantic highlighting when idle (no need to save the buffer any more)
- deleted ensime-builder functions/variables

This PR goes with ensime/ensime-server#794 but it's backward compatible so it can be merged before or after the other.

Also I'd like to know if you find any problems with running sem. highlight when idle.


rorygraves:
These changes look awesome!

aemoncannon:
Really cool. Did you also remove the explicit write-buffer calls in ENSIME completion utils?

aemoncannon:
Maybe change the ";; Requires..." comment to an assertion.

edani:
```
Really cool. Did you also remove the explicit write-buffer calls in ENSIME completion utils?
```
Whoops! I'll fix that shortly.

aemoncannon:
Hmm, are regexp and replacement reversed in ensime-buffer-contents-as-string (for the windows case)? Looks like it's going to replace line feeds with more carriage returns : )

edani:
ensime-buffer-contents-as-string takes the buffer contents from which emacs removed all the `\r`'s and adds them back.

edani:
Fixed the two issues raised by @aemoncannon

fommil:
are you happy for this to be merged @edani ?

edani:
@fommil  Yes I'm ready for this to be merged, as well as ensime/ensime-server#802 , ensime/ensime-server#803 and ensime/ensime-server#794.

fommil:
Awesome stuff, looking forward to using it this week! I'll have to untrain myself from saving so often

PR #81, opened 2015-01-23
Ensure we use correct dependency versions in integration tests.

aemoncannon:
Actually this isn't quite doing what it purports to do yet. Stay tuned.

### DIFF
--- a/ensime-pkg.el
+++ b/ensime-pkg.el
@@ -1,7 +1,7 @@
 (define-package "ensime" "0.9.10"
   "ENhanced Scala Interaction Mode for Emacs"
   '((s "1.3.0")
-    (dash "2.8.0")
+    (dash "2.10.0")
     (popup "0.5.0")
     (auto-complete "1.4.0")
     (scala-mode2 "0.21")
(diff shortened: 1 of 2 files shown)


aemoncannon:
The problem: Our tests can execute with dependency versions newer than those in ensime-pkg.el. Someone can install ensime, satisfy all declared deps, but still crash due to a stale package.

Unfortunately, I'm not sure it's possible to install a package *at* a particular version. I think we could pin at a package revision, but that seems like a pain. Ideas?

Either way, this PR is an improvement over the previous state of affairs because it pulls the package list from ensime-pkg.el.

aemoncannon:
@milkypostman Perhaps you have some advice? We'd like to test our package vs the minimum versions of all dependencies, as declared in ensime-pkg.el. As it stands, the test code runs against the latest available versions, and we could be overlooking bugs that only appear in older (but >= minimum) versions.
Is there a way to install older versions?

aemoncannon:
Merging this as it fixes the build. May follow up with another PR.

PR #82, opened 2015-01-24
Handle symlinks properly when assigning source buffers to connections. Resolves ensime/ensime-server#714

aemoncannon:


aemoncannon:
Why not use Emacs' file-in-directory-p? 
If I have a buffer open: /proj/src/X.scala,
and src is a symlink to /tmp/sources, (file-in-directory-p "/proj/src/X.scala" "/proj") will answer nil, where ensime-dir-contains-file-p will answer t.

aemoncannon:
In other words, file-in-directory-p is testing for physical containment, and ensime-dir-contains-file-p is testing for logical containment.

fommil:
cool! Can you please put the explanation you gave in the PR into the function docs? One day we might forget why we did it this way. And by we, I mean me... in about a week.

aemoncannon:
Sure thing : ) will update.

aemoncannon:
updated doc & squashed.

omnisis:
FYI, I was being affected by this bug.  Just merged aemoncannon's fix into my local ensime-emacs folder re-started emacs and I can load projects with symlinked source dirs now!

fommil:
This will help out with default scalajs cross compile projects

PR #83, opened 2015-01-25
Rename ensime-dir-contains-file-p to ensime-path-includes-dir-p. Add tests.

aemoncannon:
I feel like this name better describes what the function is actually doing.

### DIFF
--- a/ensime-config.el
+++ b/ensime-config.el
@@ -44,7 +44,7 @@
 		    (unless no-ref-sources
 		      (when-let (dir (plist-get conf :source-jars-dir))
 				(list dir)))))))
-      (-first (lambda (dir) (ensime-dir-contains-file-p file dir))
+      (-first (lambda (dir) (ensime-path-includes-dir-p file dir))
 	      source-roots))))
 
 (defmacro ensime-set-key (conf key val)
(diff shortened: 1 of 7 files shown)


fommil:
I'm running that again incase it was a transient. I don't see how this could have broken anything :confused:

aemoncannon:
Hrm, I sense it's something unrelated to my PR. I will take a look.

aemoncannon:
I can repro locally now. It only crops up with the newer protocol. Stay tuned.

Btw added elisp to dump backtraces in the event of elisp exceptions (doesn't help in this case, but might be useful later)

aemoncannon:
@fommil @edani  I think there's just one failing assertion left. In the 'expanding parameters' test. Most of this is fallout from the protocol update, which for some reason (I'm guessing order-of-PR-merging) was not caught by the test suite. Should have it cleaned up soon.

fommil:
:smile:

aemoncannon:
Thought for sure that was the last of it. Hung on formatting test just now. Hrmmm

aemoncannon:
Fixed formatting, semantic highlighting failing now

aemoncannon:
Woo! Finally. Merging.

PR #84, opened 2015-01-26
Indent region if mark is active

andrebeat:
This allows the use of the `tab` key to indent a marked region. Otherwise, we indent the current line.

### DIFF
--- a/ensime-company.el
+++ b/ensime-company.el
@@ -108,7 +108,9 @@
   (interactive)
   (when (or (ensime-at-bol-p)
 	    (not (ensime--company-try-completion)))
-    (indent-according-to-mode)))
+    (if mark-active
+        (indent-region (region-beginning) (region-end))
+      (indent-according-to-mode))))
 
 (defun ensime-company-enable ()
   (set (make-local-variable 'company-backends) '(ensime-company))


fommil:
thanks for the contribution! I'm struggling to understand when the mark would be active and we'd be completing, can you please give us an example?

BTW, I wouldn't trust the indentation to actually work according to your formatting rules. The indentation is provided by `scala-mode2` whereas ENSIME "formatting" is handled by scalariform.

andrebeat:
Currently, if I select a region and hit `tab` only the first line is indented, this change makes it indent the whole region (which used to work previously, not sure what commit broke it). I'm an emacs lisp noob so I'm not sure if this makes any sense (i.e. if the mark should ever be active when we're completing), but it seems to work appropriately.

Either way, we bind this function to tab (https://github.com/ensime/ensime-emacs/blob/master/ensime-company.el#L119) so isn't it possible for the mark to be active?

Cheers!

fommil:
aaah, ok. I see. This looks good. Thanks!

fommil:
we're having some trouble with our CI tests at the moment, we'll continue merging when things return to a healthy green state.

andrebeat:
:+1:

PR #85, opened 2015-01-27
Fix race condition when server is started but not ready

edani:
Fix ensime/ensime-server#806.

The earliest time to safely send rpc commands is after receiving the
connection info that includes the protcol version. But it turns out that
most rpc commands don't work until the analyzer is ready so we wait
for that.

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -84,29 +84,17 @@ This is automatically synchronized from Lisp.")
 (ensime-def-connection-var ensime-protocol-version nil
   "The protocol version used on the connection.")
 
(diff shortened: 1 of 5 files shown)


edani:
@aemoncannon I wasn't able to reproduce the problem you saw so this is a bit of a shot in the dark. In any case this change makes "analyzer not ready" messages go away.

aemoncannon:
@edani Nice! I think this will probably do the trick.

aemoncannon:
p.s. These are good changes that will help on my PR, and the integrations
are broken anyway, so go ahead and merge when you're ready.

PR #89, opened 2015-02-01
Use sbt to save ENSIME's classpath

edani:
Fix ensime/ensime-server#552 based on @fommil's work. 

The classpath is saved in `~/.emacs.d/ensime/classpath_<scala_version>_<ensime_version>`  once, and after that isn't touched. The user can type `M-x ensime-update` to run sbt again. I'll need to update the manual...


fommil:
Oh cool! I hadn't even thought to share globally :-)

fommil:
Should we add java version as well?

fommil:
Would it make sense to start refactoring the startup functionality into a separate file? ensime-mode in starting to get really big

fommil:
I tried this out. **instantly** started the second time... pretty funky :smile:

I noticed that the java home that was picked up by the classpath generator is incorrect. Are we correctly setting `PATH` to have the preferred `JAVA_HOME/bin` at the front?

fommil:
@edani I've sent you a PR that solves the `JDK_HOME/lib/tools.jar` issue.

The only remaining comment I have is that we should be including the JDK path in the file. I have no idea how to do that cleanly. Maybe a hashcode of the java pathname?

This is looking really really sweet and should make a massive difference to start responsiveness, particularly for people with slow/no internet connection and low-end machines.

edani:
@fommil About JAVA_HOME and the Java version... How about we don't save `lib/tools.jar` in the classpath file at all, and prepend it to the classpath in `ensime--start-server`

edani:
@fommil Actually I'm not sure what JDK_PATH does. How does it affect the result of `ensime--update-server` ?

fommil:
That could work! 

JDK_HOME is used inside the scala stdlib to detect the jdk, so we set it to avoid conflicts with user preferences. It's not needed to run the sbt script.

fommil:
one other thing... can we automatically run `ensime-update` when we detect that the `ensime-mode.el` file (or whatever it is refactored into) is more recent than the classpath file?

That way, when the user updates ensime in MELPA, the server will get updated as well, which I think will avoid a lot of protocol incompatibility changes.

Or is there a hook that we can run when installing from MELPA?

edani:
@fommil  I think we can do the first thing (checking the timestamp on .el files). As far as updating the server when MELPA runs, that would be hard because we don't have access to the desired scala version.

I'll send a new commit that addresses all your comments.

fommil:
@edani cool! Don't forget to look at my PR to your branch as it'll save you implementing some of the fixes.

edani:
Done, I'll squash if you;e happy with the changes.

fommil:
I created a fix for my comment above at gave you a PR https://github.com/edani/ensime-emacs/pull/2

Please do squash (into a commit with your name against it :smile:) this is a really fantastic change. Is it working ok in windows?

edani:
Merged @fommil's fix and squashed the whole lot.

fommil:
restarted travis incase it was transient

fommil:
same again. Any ideas why the CI is failing?

fommil:
I don't think the CI is failing because of this change, I've been using it all week and good to hear Windows confirmation.

:smile: this is a really really useful update

@edani / @aemoncannon can one of you please look at the emacs build and see why its failing?

aemoncannon:
@fommil @edani  It kinda does look related. Btw we get stack traces in the emacs travis log now. Here's the relevant portion after a little cleanup. Looks like (ensime--get-cache-dir config) is returning nil.

Starting test: 'Test inspect type at point.'
Updating ENSIME server...
Debugger entered!
  (file-name-as-directory nil)
  (let* ((config (ensime-config-load config-file)) (root-dir (ensime--get-root-dir config)) (cache-dir (file-name-as-directory (ensime--get-cache-dir config))) (name (ensime--get-name config)) (scala-version (plist-get config :scala-version)) (server-env (or (plist-get config :server-env) ensime-default-server-env)) (buffer (or (plist-get config :buffer) (concat ensime-default-buffer-prefix name))) (server-java (file-name-as-directory (plist-get config :java-home))) (server-flags (or (plist-get config :java-flags) ensime-default-java-flags))) (make-directory cache-dir (quote t)) (let* ((server-proc (ensime--maybe-start-server (generate-new-buffer-name (concat * buffer *)) server-java scala-version server-flags (cons (concat JDK_HOME= server-java) (cons (concat JAVA_HOME= server-java) server-env)) config-file cache-dir)) (host 127.0.0.1) (port (function (lambda nil (ensime-read-swank-port (concat cache-dir /port)))))) (display-buffer-at-bottom (process-buffer server-proc) nil) (process-put server-proc :ensime-config config) (setq ensime-server-processes (cons server-proc ensime-server-processes)) (ensime--retry-connect server-proc host port config cache-dir)))

fommil:
oops, maybe I shouldn't have merged so quickly

edani:
I'll take a look. I had run tests at some point but I skipped them for the last PR... shame on me!

PR #90, opened 2015-02-10
Fix test suite

edani:
Not sure if travis has JAVA_HOME set, we'll see.


fommil:
nice, it's great to have these tests.

aemoncannon:
+1, thanks for fixing

PR #92, opened 2015-02-11
Cache known bad connections. A(nother) candidate fix for ensime/ensime-server#822

aemoncannon:


aemoncannon:
@fommil This *should* take care of the issue.

fommil:
cool, thanks that makes a big difference! :smile:

aemoncannon:
Cool

PR #93, opened 2015-02-16
Use java_home cmd to find JAVA_HOME

aemoncannon:


aemoncannon:
Little fix so things work on osx before I look at @fommil's protocol stuff.

PR #98, opened 2015-02-25
Wire up elisp to browse documentation at point.

aemoncannon:


aemoncannon:
btw, is bound to C-c C-v d

PR #105, opened 2015-03-03
enable emacs 24.4

fommil:


fommil:
coverage decreased because I removed an exclusion rule.

PR #106, opened 2015-03-03
miscellaneous test changes

fommil:


fommil:
I have no idea why coverage is coming back at 3%.

**sigh** it's all borkened

aemoncannon:
@fommil Btw, I noticed it looks like undercover sends it's results in th kill-emacs-hook, i wonder it that could be related somehow. Looks like they are careful to (ignore-errors ...), though.

fommil:
Hmm. It's definitely not related to the emacs 24 problem because we don't use undercover there, but that's worth investigation. These test failures are driving me mad, I wonder if travis just isn't powerful enough and we need to limit the JVM resources.

fommil:
yay! Coverage seems to be working again. I'll tidy up some loose ends and self merge when done, this is all test infrastructure boring stuff.

PR #109, opened 2015-03-08
enable emacs version matrix

fommil:


fommil:
yay! Coverage is now seeing everything.

aemoncannon:
Awesome work!

fommil:
I'm thinking about doing  a bisection on the emacs source code to try and work out when they introduced the return value regression.

aemoncannon:
Yeah. Great candidate for bisection...

PR #110, opened 2015-03-13
Fixes ensime/ensime-server#815: close refactoring buffer after close.

alpeb:
Note other 'popup' buffers not related to refactoring also suffer from this, but this commit only addresses those for refactoring


fommil:
nice! :smile:

fommil:
i bet that's a transient. it happens when the travis machines are under load.

fommil:
I'm checking this locally, if it passes, I'll merge.

PR #111, opened 2015-03-13
small test improvements

fommil:


aemoncannon:
Do we still get a backtrace this way?

fommil:
@aemoncannon yeah, but it could probably do with some nicer formatting. I was actually getting a hang with the original (not in this project, but when I stole it and used it when validating my `.emacs` file).

PR #112, opened 2015-03-14
Test improvements

fommil:


fommil:
this needs server support to pass, so merging early

PR #117, opened 2015-03-18
Auto-generate .ensime

edani:
Fix ensime/ensime-server#773.   When an sbt project is detected, run
"sbt gen-ensime" before startup.  Feature is off by default and
controlled by 'ensime-auto-generate-config'

Fix ensime/ensime-server#695

### DIFF
--- a/ensime-startup.el
+++ b/ensime-startup.el
@@ -108,15 +108,16 @@
 
 (defun ensime-reload ()
   "Re-initialize the project with the current state of the config file.
-Analyzer will be restarted. All source will be recompiled."
+Analyzer will be restarted."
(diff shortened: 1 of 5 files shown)


ensime-startup.el
```diff
@@ -18,6 +18,56 @@
       (error "Name in ensime configuration file appears to be unset"))
     name))
 
+
+(defun ensime-refresh-config ()
```
aemoncannon: Maybe move these functions into ensime-config?

aemoncannon:
lgtm

edani:
Rebased and moved most code to ensime.config.el

PR #120, opened 2015-03-22
Use new RPCs to make inspector more accurate.

aemoncannon:
Blocked on corresponding server PR.


fommil:
can you please rebase with upstream? That will use the shippable stuff. I'm guessing there are no tests to cover the new behaviour?

I don't know why travis is kicking off. I thought I stopped it.

PR #122, opened 2015-03-24
Add tests of the type inspector UI

aemoncannon:


aemoncannon:
Looks like our old friend 'spurious non-zero status'...

fommil:
hmm, but on all the branches?

aemoncannon:
so confused by this: we print status is 1, which means that
ensime--test-had-failures must be true, but how can
ensime--test-had-failures be true if no test failed? something weird going
on.

aemoncannon:
Working theory: there *is* a failing test, but somehow we're not logging it.

fommil:
I've made the build matrix only for two versions of emacs, can you rebase please?

PR #124, opened 2015-03-27
Add a regex for matching scala idents in reverse. 

aemoncannon:
This prevents worst-case behavior in forward matching that could (appear to)hang emacs. Resolves ensime/ensime-server#914.


fommil:
btw, I think the build is broken because of the github DDOS

PR #125, opened 2015-03-28
Reuse the index directory in the slow test suite

edani:
Also fix a couple Windows problems

Fix ensime/ensime-server#892

I'm pretty happy with the speed improvement as it is, but maybe the next thing we can try is reuse the server between tests, resetting the pres compiler each time.

### DIFF
--- a/ensime-test.el
+++ b/ensime-test.el
@@ -67,9 +67,8 @@
     (insert-file-contents
      (ensime--classpath-file ensime--test-scala-version))
     (--first
-     (string-match (concat "/scala-library.*\.jar") it )
+     (string-match "[/\\]scala-library.*\\.jar$" it )
      (split-string (buffer-string) ensime--classpath-separator 'omit-nulls))))
-
 
 
 (defun ensime-test-concat-lines (&rest lines)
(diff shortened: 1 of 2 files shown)


fommil:
@edani awesome! BTW, I'm experiencing shippable problems and can't see the builds. Can you see them?

fommil:
maybe we should have a minimal Windows virtual image that we run the tests on :smile: you seem to be forever fixing our regressions....

rorygraves:
Lol. You are getting good at that @fommil

aemoncannon:
I also can't see shippable builds. Since last night

edani:
I can't see the build either. Could be because my shippable user can't see the ensime organization.

fommil:
they said there is a regression in the github API. If you enable private repos in Shippable, it starts to work again.

BTW, Emacs tests are all timing out. I wonder if we merged a regression this week.

fommil:
ok, I figured out what it is. You made it too fast :stuck_out_tongue: 

Let's take "Test inspect type at point" as an example:

```elisp
    ((:full-typecheck-finished val)
     ... ignore ...)

    ((:indexer-ready val)
     ... do test...)
```

and normally that's fine because we get the typecheck first and then we do the work when the slow indexer is ready.

But **now** the indexer arrives immediately, and the test never runs.

I think there is a long overdue refactor of the tests needed here. Basically all the async tests need to wait for

```elisp
(:connected connection-info)
(:full-typecheck-finished val)
(:indexer-ready val)
```

before they can start, so perhaps we should have a triggering criteria that those things arrive and then we start the test.

BTW, I reduced the number of branches in the build to two on master to check that we didn't introduce a regression without realising. It certainly speeds up the build, so perhaps you could rebase onto that and when Shippable allow me to add my dedicated host I'll add it and enable all four branches again.

fommil:
Re: additional performance improvements. I'm actually really happy to see the tests written this way, because it stresses a bunch of integration factors that we just can't test in the scala tests.

aemoncannon:
going to hack on consolidating those events - let me know if someone else
already working on that..

aemoncannon:
Actually, maybe you wanna tackle it in this PR,  Eric, to avoid the
conflicts? I was thinking it might be cool if we could wait on an unordered
set of events. Like:

((await-all :connected :full-typecheck-finished :indexer-ready)
 ... do test ...
)

rorygraves:
@aemoncannon I like the await-all - I should update the scala tests to do the same.

rorygraves:
It is funny that we have issues with our tests because we make them too darn fast though ;)

edani:
OK, I'll work on a "await-all" type of control. It's strange though, in my PR I switched the order of :full-typecheck-finished  and :indexer-ready for this very reason.

edani:
There's something else going on. According to the build log, the `:indexer-ready` event never arrives.  I'll try and run the tests on linus manually to see if I can figure it out.

fommil:
that's extremely concerning and would be a server bug.

edani:
The last shippable buildpassed but I still can't explain earlier failures. I'm also having trouble reproducing them locally.

rorygraves:
Don't forget we have rolled back one of my startup changes, I'm not sure of the crossover timings, but could be related.

fommil:
awesome! This will test a code path that is quite common: opening an existing project :smile:

fommil:
wow, cross build integration tests in 7 minutes (including time to dist). Truly we are living the dream.

fommil:
I might add the other two versions of emacs back into the mix.

PR #126, opened 2015-04-02
use maven resolver for ENSIME server resolution

fommil:


fommil:
I noticed that the ensime-server integration tests are downloading a lot of jars again, it was because we use the maven resolver in ensime-server but the default ivy resolver in the client. This reconciles. Hopefully upstream sbt will fix this once and for all and make it the default.

fommil:
this is a trivial change and seems to have fixed the issue of downloading (which isn't functional, more an optimisation)

PR #128, opened 2015-04-11
Allow elisp to be byte compiled

edani:
This fixes ensime/ensime-server#310

- Move all macros to ensime-macros.el
- Fix dependency problems
- Fix some errors pointed out by byte compilation

I tested it with a local copy of melpa.  In the future we just need to keep all macros in ensime-macros.el file or we risk creating more errors.

At some point I'd like to fix every byte-compiler warnings: they're useful to detect typos and other problems. But there are so many spurious warnings right now that they mask the important ones.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 .#*
 *.*#
+*~
 
 TAGS
 emacs.d
+*.elc
--- a/ensime-auto-complete.el
+++ b/ensime-auto-complete.el
@@ -19,7 +19,12 @@
 ;;     Software Foundation, Inc., 59 Temple Place - Suite 330, Boston,
 ;;     MA 02111-1307, USA.
 
+(eval-when-compile
+  (require 'cl)
+  (require 'ensime-macros))
 (require 'ensime-completion-util)
+(require 'ensime-model)
+(require 'ensime-util)
 (require 'auto-complete)
 
 (defcustom ensime-ac-enable-argument-placeholders t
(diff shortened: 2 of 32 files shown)


aemoncannon:
great work. I wonder if this will help with the coverage bug i was seeing...

edani:
This triggered the timeout problem again. Here is one clue, this looks like a race condition:

```
Reply to canceled synchronous eval request tag=ensime-result-8-sym69488 sexp=(swank:completions (:file "/tmp/ensime_test_proj_153260/src/main/scala/hello_world.scala" :contents "package com.helloworld
import scala.collection.immutable.List
import scala.collection.immutable.{ List, Vec}
class HelloWorld{
import sc
}") 136 30 nil t)
```

Also it turns out that compiling *.el before the test suite prevents undercover from doing its job. We may need to run the test suite twice, once for undercover and once to test compiled elisp.

rorygraves:
Good stuff

aemoncannon:
hypothetically, if the elisp compiles without errors, it should do the same
thing as uncompiled elisp. so maybe we could just compile as a lint step,
then throw out the bytecode and run tests

fommil:
this is looking good and passing the tests (I kicked off a re-run, it seems we do still have some transients left). Can you please squash and then merge if you're also happy with it.

This was even has a bounty! :smile:

aemoncannon:
awesome!

fommil:
cool, I think this actually speeds up my scala project start time by a few seconds.

PR #130, opened 2015-05-03
Fix potential race conditions in tests

edani:
This is something we talked about a few weeks ago. Right now I don't think there are any race conditions, but these changes will prevent them from happening in new tests.


fommil:
fantastic! This will really help the build stability

aemoncannon:
really cool. this makes the tests a better (more accurate) doc for the protocol as well.

PR #131, opened 2015-05-11
use correct version of scala in scalap and scalac

fommil:
note this allows typelevel scala to be used in the user's projects


fommil:
debug tests timed out again (transient). They are still a little fragile.

PR #133, opened 2015-05-28
defined stackframes can be folded

kleewho:
I tested it after pasting stacktrace to buffer that was created by `ensime-stacktrace-switch`. I realize that it's not finished but I would rather like to get some hints about what should I change after someone have a chance to look at code.

This should fix https://github.com/ensime/ensime-server/issues/566 eventually

### DIFF
--- a/ensime-stacktrace.el
+++ b/ensime-stacktrace.el
@@ -4,14 +4,25 @@
   (require 'cl)
   (require 'ensime-macros))
 
+(require 'fold-this)
+
 (defconst ensime-stacktrace-buffer-name-base "*ensime-stacktrace*")
 
 (defvar ensime-stacktrace-buffer-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-c C-c") 'ensime-stacktrace-highlight)
+    (define-key map (kbd "C-c C-f") 'ensime-stacktrace-fold-buffer)
+    (define-key map (kbd "C-c C-p") 'fold-this-unfold-at-point)
     (define-key map (kbd "C-c C-q") 'quit-window)
     map)
   "Keymap for `ensime-stacktrace-buffer-mode'.")
+
+(defvar ensime-stacktrace-foldable-stackframes '("at akka\\.*"
+                                                 "at scala\\.*"
+                                                 "at java\\.*"
+                                                 "at jline\\.*"
+                                                 "at sbt\\.*"
+                                                 "at xsbt\\.*"))
 
 (define-minor-mode ensime-stacktrace-buffer-mode
   "Mode for highlighting stack traces"
@@ -40,6 +51,8 @@
          (when (= (buffer-size buf) 0)
            (insert ";; Stack trace buffer\n")
            (insert ";; Paste a stack trace below and press `C-c C-c' to create links to source code.\n")
+           (insert ";; Press `C-c C-f' to fold all defined stackframes\n")
+           (insert ";; Press `C-c C-p' to unfold region that you're currently on\n")
            (insert ";; Press `C-c C-q' to leave this buffer\n")
            (insert "\n"))
         (local-set-key (kbd "C-c C-c") 'ensime-stacktrace-highlight)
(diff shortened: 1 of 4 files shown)


fommil:
tests are failing due to

```
Done (Total of 4 files compiled, 2 skipped in 4 directories)
(file-error Cannot open load file fold-this)  eval-buffer
```

fommil:
ok, I see why that fails. You need to add `fold-this` as a MELPA dependency in the package definition file `ensime-pkg.el`. Should be an easy pattern to follow in there.

kleewho:
@fommil I'll fix it today. I didn't look much into source code, I was just focusing on this one file.

fommil:
you might want to edit the help text that appears at the top of the stack trace buffer to say "use commands foo and bar to fold/unfold regions" And maybe set up some canonical `fold-this` keybindings for this mode.

kleewho:
Yeah, that's the kind of feedback I wanted to have, because I'm not using ensime that much. I will do that of course, thank you!

fommil:
I invoked `ensime-stacktrace-fold-buffer` but got

```
let: Symbol's value as variable is void: whatever-stacktraces-to-foldInvalid face reference: compilation-error-face
```

fommil:
I think you might need to be escaping your regex: `"java.*"` => `"java\\.*"` as dot is interpreted as "any char".

kleewho:
@fommil, I had very dirty version with naming like: 'whatever', 'bla-bla' and then I was renaming stuff but I missed that one. I should test it after fresh start just to be sure!

fommil:
maybe `C-c C-f` to fold everything (documented) and `C-c C-p` to unfold at point? (open to alternative suggestions).

fommil:
oooh, it would be awesome to be able to auto fold not only stacktraces but entire threads that match a pattern by name or if all their stackframes is folded. Maybe for v2  :smile:

kleewho:
@fommil I'm open to suggestions! start thinking about v3 ;)

fommil:
classic example of a boring thread

```
"BoneCP-pool-watch-thread-ESDatabase (jdbc_h2_mdes  XXXXXXXX)" #32 daemon prio=5 os_prio=0 tid=0x00007f26d3e95800 nid=0x452c waiting on condition [0x00007f25e3ffe000]
   java.lang.Thread.State: WAITING (parking)
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000004048f4d60> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
	at java.util.concurrent.ArrayBlockingQueue.take(ArrayBlockingQueue.java:403)
	at com.jolbox.bonecp.PoolWatchThread.run(PoolWatchThread.java:75)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)

   Locked ownable synchronizers:
	- <0x0000000405ee01f8> (a java.util.concurrent.ThreadPoolExecutor$Worker)
```

kleewho:
I can't find build log to see whether it failed once again because of missing `fold-this.el` in package definition. On shippable I get: 

> The build you're looking for wasn't found.

fommil:
heh, typo `(void-function defime-key)  let`

You know you can run the tests locally? The `shippable.yml` file shows exactly what is running.

kleewho:
I'll take look. I didn't know that.

kleewho:
@fommil Do you want me to write tests for this new thing? I see that the coverage didn't drop in % cause `ensime-stacktrace` wasn't tested at all but I added some untested lines there.

fommil:
unit tests are always good

fommil:
Fails with

```
command-execute: Symbol's function definition is void: ensime-stacktrace-buffer-map
Invalid face reference: compilation-error-face [40 times]
```

(the missing face could be erroneous)

kleewho:
@fommil I think you have more problems with this then the added value ;) Another _typo_ I would say - this time using company completion to eagerly. To make it up to you I'll fix all those crappy bugs I made and I'll start thinking how to collapse nicely those boring threads

fommil:
when I do `C-x C-f` all I get is
```
Mark set
```

and then a list with three numbers in it.

ensime-stacktrace.el
```diff
@@ -40,12 +47,23 @@
          (when (= (buffer-size buf) 0)
            (insert ";; Stack trace buffer\n")
            (insert ";; Paste a stack trace below and press `C-c C-c' to create links to source code.\n")
+           (insert ";; Press `C-c C-f` to fold all defined stackframes")
```
fommil: need `\n` newlines on these two lines

kleewho:
@fommil After work I restart my emacs and I'll check what's going on in fresh emacs env. I thought that I just did that last time, but apparently didn't. And all my fixes when in work seems to not fix anything cause I'm to distracted. So I'll let you know in 8-10h from now if it's worth checking or I have to fix something first.

fommil:
btw, there is a typo in the face definition... can you please rename `compilation-error-face` => `compilation-warning` (warning being more appropriate than error here) while you're in here?

kleewho:
@fommil ok, I'll fix this as well

kleewho:
@fommil can you give me an example that you're using as a test?

fommil:
of a stacktrace? just start up ensime and take a stacktrace with the `jstack` tool. Actually, it would be pretty neat to be able to populate the buffer with a call out to `jstack` - you did ask for v3 suggestions :wink:

fommil:
(btw, I don't know why coveralls is being so slow)

kleewho:
@fommil any help with the time out? I can't see any details on shippable

fommil:
Neither can I. Unfortunately shippable has become as trustworthy as a drunk Frenchman in a brothel. I'll try this out again tomorrow, I didn't have to opportunity today.

fommil:
damnit, sorry I totally forgot about this today! Actually my work project is currently in refactor and is rarely building so I don't think I had an ensime server up and running at all.

Shippable, you are just crap.

fommil:
fantastic! This is really great. A few small suggestions before final acceptance:

1. fold by default?
2. when folding everything, make sure to not apply a fold twice or `...`s all add up
3. add `jline.*` and `sbt.*` and `xsbt.*` to the list to fold

kleewho:
I was trying to address 2. and 3. but it was late so there might be dragons there. I have to think how to achieve 1. in a way that won't drive users nuts.

fommil:
```
list: Lisp nesting exceeds `max-lisp-eval-depth'
```

:sob:

kleewho:
heh... I told you that dragons might be there. Probably I forget  `)` somewhere. Advice to myself, don't code when it's late and you're back from the bar.

kleewho:
I looked at the file briefly and it might be something different then I thought

kleewho:
@fommil should be better. Please check against another long stacktrace ;)

kleewho:
@fommil I'm not sure that I can trust shippable:

>  The build you're looking for wasn't found.

fommil:
Sorry I'm travelling, but I'll definitely look at it!

fommil:
nice! Screw shippable. Some small tweaks might be:

- `C-c C-c` folds
- TAB unfolds at point

PR #136, opened 2015-05-31
ensime--retry-connect: after successful connection, don't try to dele…

trollusk:
ensime--retry-connect: after successful connection, don't try to delete the current window if it's the only window, as doing this creates an error which breaks ensime.

Instead, switch the view back to the previous buffer, which is usually the buffer from which 'M-x ensime' was invoked.


fommil:
Cool. Presumably this is for the case where you launch ENSIME then close the launching buffer

PR #137, opened 2015-06-05
hopefully work on drone

fommil:


fommil:
blocked by https://github.com/drone/drone/issues/1048

fommil:
@edani @aemoncannon I've very nearly got the emacs tests working in `-batch` mode, which means they will run in drone and won't need a fake display anymore!

But I hit this:

```
error in process filter: Window #<window 44 on *ensime-search-results*> too small for splitting 5
```

in my last run. Any ideas? http://fommil.com/github.com/ensime/ensime-emacs/drone/012b8ad32e065847a48896db39109c0921ab1109

actually, on a somewhat related note... I wonder if perhaps we couldn't just feed IDO or something with our search results, instead of having our own custom mode for it. It feels like there should already be a view out there for displaying live search hits.

fommil:
I think temporarily disabling the interactive search test is the lesser of two evils. I'll raise a ticket to get them working again.

PR #138, opened 2015-06-08
Don't prompt for config filename when it's known

greenrd:
Fixes ensime/ensime-server/#937

### DIFF
--- a/ensime-startup.el
+++ b/ensime-startup.el
@@ -100,6 +100,10 @@ saveClasspathTask := {
           (ensime--update-server scala-version `(lambda () (ensime--1 ,config-file)))
         (ensime--1 config-file)))))
 
+(defun ensime--maybe-update-and-start-noninteractive ()
+  (let ((ensime-prefer-noninteractive t))
+    (ensime--maybe-update-and-start)))
+
 (defun* ensime--1 (config-file)
   (when (and (ensime-source-file-p) (not ensime-mode))
     (ensime-mode 1))
--- a/ensime.el
+++ b/ensime.el
@@ -81,7 +81,7 @@
 (defvar ensime-protocol-version "0.7")
 
 (defvar ensime-prefer-noninteractive nil
-  "State variable used for regression testing.")
+  "State variable used for regression testing, and for skipping prompt in conjunction with sbt.")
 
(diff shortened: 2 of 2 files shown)


fommil:
Thanks! You could have bloody waited until after scala days, now I have to update the contributor list :-P

Did you run the tests locally? I have zero faith in the shippable build and I plan to replace it with drone immediately.

fommil:
I just tried this out... I opened up my project base directory with `dired` and typed `M-x ensime` but it still prompted me for the config file. Is that expected?

greenrd:
You have to customize `ensime-auto-generate-config` to `t`. This pull request only covers cases when the config is auto-generated.

fommil:
@greenrd what do you think about making the auto discovery independent of the auto-generated sbt? I'd quite like this feature, but not the auto generate. A separate `defvar` for the option, on by default.

fommil:
I like this but I'd like to get the "default to autodetected file" working for starting the server/REPL if thats possible. See the linked PR for more context.

fommil:
@greenrd are you happy to make the requested changes here or in a follow up?

greenrd:
@fommil yes - I've been busy moving house recently, but I'll try to look at it this weekend.

fommil:
no problem, take a few weeks its not urgent :smile:

fommil:
ping

fommil:
btw, just incase you weren't aware, you can get the config for the current buffer by doing this

```
  (let* ((connection (ensime-connection))
         (config (ensime-config connection))
  ...)
```

it won't work for non-sources, but I don't think it should be too hard to fix that (might involve a hack involving looking at the `:root-dir` after all source checks fail).

fommil:
Ping

greenrd:
I'm not going to have a chance to look at this until I go on holiday next Tuesday.

fommil:
@greenrd new houses, holidays?! Anybody would think you had a life outside scala :-)

No problem, I'm just doing the pinging rounds on all open PRs. Leave it until after your hols.

fommil:
@greenrd ! ping

btw, somewhat relevant is #253

fommil:
I'm cherry picking this onto a PR that I hope to raise tonight.

PR #139, opened 2015-06-13
Use hideshow instead of fold-this for foldable stack frames.

jeberger:
Hideshow is included with emacs but fold-this isn't, so using hideshow makes it easier to install

### DIFF
--- a/ensime-pkg.el
+++ b/ensime-pkg.el
@@ -8,7 +8,6 @@
     (yasnippet "0.8.0")
     (popup "0.5.0")
     (scala-mode2 "0.21")
-    (fold-this "0.3.0")
     ))
 
 ;; Local Variables:
--- a/ensime-stacktrace.el
+++ b/ensime-stacktrace.el
@@ -4,7 +4,8 @@
   (require 'cl)
   (require 'ensime-macros))
 
-(require 'fold-this)
+(require 'hideshow)
+(require 'dash)
 
 (defconst ensime-stacktrace-buffer-name-base "*ensime-stacktrace*")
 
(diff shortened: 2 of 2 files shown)


fommil:
cool! But I don't get any visual indication when something is folded... and `C-c C-p` doesn't seem to unfold anything. Can you look into that please and we'll merge when it's fixed.

jeberger:
This should fix it...

fommil:
I'm not seeing anything. I'm also not able to run `C-c C-p` to expand anything.

Are you testing this after doing a `C-c C-c`?

jeberger:
Yes, `C-c C-c` to parse the stack trace, then `C-c C-f` to enable folding and `C-c C-p` to expand the folded parts. Some points worth checking:

 - Folding does not work as-is. I had to manually `(require 'dash)` and to change `--any` to `--any?` on line 84 of `ensime-stacktrace.el`. This issue is also present in the original code. I haven't done anything about it since I assumed that the original works for other people, but you might want to check if you are affected.

 - Folding leaves the first line of the folded section visible with `...` appended, unless the section consists only of a single line in which case folding doesn't do anything.

 - `C-c C-p` only expands the section pointed to by the cursor. You need to have the text cursor somewhere on the same line as the `...` but not at the end of the line for it to expand.

 - Standard hideshow shortcuts still apply, so `C-c @ C-s` should also allow you to expand a single section and `C-c @ C-M-s` should expand everything.

fommil:
hmm, can you add your dash fixes to the PR? Maybe that'll help

fommil:
btw, ignore Shippable. It's broken.

fommil:
nice! That works for me. Can you please squash your commits?

jeberger:
How do I do that? I ususally use Mercurial and am not familiar with the git commands. I tried `git rebase -i` which allowed me to squash the commits locally, but now github is rejecting my push…

fommil:
when you rewrite history, you need to force push (or `sith` as I call it :stuck_out_tongue: ) `git push -f`

jeberger:
Thanks, done.

fommil:
thanks! :smile:

PR #140, opened 2015-06-20
Implicit conversion highlighting

edani:
This PR goes with ensime/ensime-server#1004 . I could use some feedback on the UI.

- Implicit conversions are shown using the compiler-warning face, and with a question mark in the margin
- For implicit parameters, only the last character is highlighted. The reason is that the conversion ranges are usually pretty large (think

```
something map {
...
}
```

so I didn't want to highlight the whole thing

- hovering on a highlighted area, or typing `C-c C-v e`  (same keybinding as for showing compiler notes) anywhere on a line shows all implicits that are applied on that line.

Future work should include jumping to the definition of the implicit(s) being applied.

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -879,7 +879,7 @@ copies. All other objects are used unchanged. List must not contain cycles."
 
 (defun ensime-rpc-doc-uri-at-point (file point)
   (ensime-eval
-   `(swank:doc-uri-at-point ,file ,point)))
+   `(swank:doc-uri-at-point ,file ,(ensime-externalize-offset point))))
 
 (defun ensime-rpc-doc-uri-for-symbol (fqn &optional member-name member-signature)
   (ensime-eval
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -1115,8 +1115,17 @@ copies. All other objects are used unchanged. List must not contain cycles."
   (ensime-eval `(swank:shutdown-server)))
 
 (defun ensime-rpc-symbol-designations (file start end requested-types continue)
+  (when (version< (ensime-protocol-version) "0.8.15")
+    (setq requested-types (remove 'implicitParams requested-types))
+    (setq requested-types (remove 'implicitConversion requested-types)))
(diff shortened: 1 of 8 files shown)


fommil:
I'm seeing this every time I save a file:

```
Error running ensime-source-buffer-saved-hook: (wrong-type-argument listp +)
```

fommil:
I'm noticing a lot of stray error/warning markers in my sidebar. Has the logic to clear them when we receive scala-notes changed perhaps? (I might have introduced a regression with my akka PR)

aemoncannon:
They should be cleared explicitly on receipt of a clear-all event...

edani:
I pushed changes to the UI: colors are less warning-y, and I added types in the implicit info. Let me know how you like it.

fommil:
Looks like two real failures

```
Error executing test: (ensime-test-interrupted . Test interrupted: (ensime-test-assert-failed . Expected truth of (memq sym-type (ensime-sem-high-sym-types-at-point)).).), moving to next.
```

```
Error executing test: (ensime-test-interrupted . Test interrupted: (ensime-test-assert-failed . Expected (ensime-implicit-notes-at (point)) to equal ("Implicit parameters added to call of stringToB(\"xxx\"): (zz: scala.Int)" "Implicit conversion of \"xxx\" using stringToB: (s: String)(implicit x: Int)pack.B"), but was nil.).), moving to next.
```

fommil:
Fantastic! I'll probably customise the face colour in `darcula` but this is nice.

Any chance of using the shortnames for the types and also changing the formatting a little? e.g.

```
Implicit conversion of sender() using actorRef2Scala: (ref: akka.actor.ActorRef)akka.actor.ScalaActorRef
```

goes to

```
Implicit conversion of sender() using actorRef2Scala(ref: ActorRef): ScalaActorRef
```

edani:
The test failure is normal, it needs the server-side change! I'll re-run the test once the server part is merged.

I'll fix the function description to use shortnames as well.

fommil:
I'll raise this with your commits rebased and merge, awesome!

PR #143, opened 2015-06-27
Utilities for importing package contents into the REPL.

dangerousben:

PR #144, opened 2015-06-27
Require config file but not server to launch a REPL.

dangerousben:
As suggested here: https://github.com/ensime/ensime-server/issues/972#issuecomment-104881777

Implementing this now because it will make it easier to write tests for https://github.com/ensime/ensime-emacs/pull/143

### DIFF
--- a/ensime-config.el
+++ b/ensime-config.el
@@ -80,8 +80,8 @@
 		  guess
 		  nil
 		  (if guess (file-name-nondirectory guess) "")))))
-
-    (if (and (file-exists-p file)
+    (if (and file
+             (file-exists-p file)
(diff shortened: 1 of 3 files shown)


fommil:
great! So this looks like it'll work if someone calls it programmatically but what about if somebody calls it interactively without a server up? Will it prompt for the `.ensime` file?

Also, note the work that @greenrd is doing in https://github.com/ensime/ensime-emacs/pull/138 as it sounds like that "default to autodiscovered" behaviour might also be appropriate here.

dangerousben:
Yes, it prompts.  If a flag is added to autodiscover independent of autogenerating the config file I expect it would just work.

PR #145, opened 2015-06-27
Added faces for deprecated symbols

ale64bit:
Added the face support for corresponding PR in `ensime-server`.

### DIFF
--- a/ensime-test.el
+++ b/ensime-test.el
@@ -1847,8 +1847,16 @@
                                  "}"
                                  "class B {}"
                                  "class C{"
-                                 " implicit def stringToB(s: String) = new B"
-                                 " val x: B = \"xxx\"/*13*/"
+                                 "  implicit def stringToB(s: String) = new B"
+                                 "  val x: B = \"xxx\"/*13*/"
+                                 "}"
+                                 "class D {"
+                                 "  @deprecated(\"BadClass\", \"1.0\")"
+                                 "  class BadClass {}"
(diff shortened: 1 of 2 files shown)


fommil:
I've kicked off a new build, new server should be available now.

fommil:
ok, this is passing the tests.

Can you please squash? And it would be great if you could update the existing semantic highlighting test to include this. The emacs tests are not just for the emacs code, it's also useful to have additional integration tests here.

ale64bit:
@fommil Done. Sometimes I get this error when running the tests:
```
ensime-macros.el: `flet' is an obsolete macro (as of 24.3); use either `cl-flet' or `cl-letf'.
ensime-inf.el: (lambda (x) ...) quoted with ' rather than with #'
ensime-goto-testfile.el: `labels' is an obsolete macro (as of 24.3); use `cl-labels' instead.
ensime-notes.el: `labels' is an obsolete macro (as of 24.3); use `cl-labels' instead.
(invalid-read-syntax ))  load-with-code-conversion
```
What about it? 

P.S.: Sorry for inexperience; I started using emacs a week ago.

fommil:
we need some code cleanup by the sounds of it

fommil:
Interesting, I think you might have introduced a regression because I've never seen this error before

```
(invalid-read-syntax )) load-with-code-conversion
```

It's so weird that I suspect a typo, extra or missing bracket somewhere.

ale64bit:
@fommil oops, sorry about that. Fixed and tested.

PR #147, opened 2015-06-28
Events to make it easier to test REPL interactions.

dangerousben:


dangerousben:
@fommil are there instructions for replicating the shippable build locally somewhere?  I've built a container from ensime-docker but now I'm a bit stuck.

fommil:
yeah, it sucks. They run some proprietary scripts. I'm hoping to move us over to drone, which is both conceptually easier and all open source.

It should be just as simple as cloning the repo and running the shell commands in the `shippable.yml` file.

fommil:
I'm seeing this on shippable

```
Starting test: 'REPL without server.'
Copying /tmp/ensime_test_proj_226vJf/cache to /tmp/ensime_test_proj_226JzH/cache
(void-function string-suffix-p) error in process filter: or
```

aemoncannon:
string-suffix-p was introduced in 24.4 -- could that be the cause?

fommil:
of course, we matrix test back to 24.1

fommil:
maybe there is something in dash?

aemoncannon:
s.el has s-ends-with?

fommil:
sorry, I didn't realise you'd rebased.

PR #189, opened 2015-07-05
comments on connection-info

fommil:


ensime-client.el
```diff
@@ -1115,6 +1115,10 @@ copies. All other objects are used unchanged. List must not contain cycles."
   (ensime-eval `(swank:shutdown-server)))
 
 (defun ensime-rpc-symbol-designations (file start end requested-types continue)
+  (while (not (ensime-protocol-version))
+    ;; workaround startup timing issue
+    (message "Waiting for connection-info")
+    (sit-for 1))
```
aemoncannon: Will the server break if we request types it doesn't know about?

fommil: That I don't know. I assume the format changed at some point. Happy birthday BTW :smile: I don't know how but my phone seems to know these things

aemoncannon:
Hey thanks : )   

@edani Do you know why the version checking is necessary?

fommil:
Oh I just spotted what it is... it's the symbol types he request. Older servers don't know about implicits and deprecation warnings.

aemoncannon:
Possible to ignore on server or is that hard with to do with the new
protocol magic?

fommil:
@aemoncannon I'm sure we probably could ignore on the server. The problem is that the older servers don't like it. We should maybe change the API so that you don't have to tell the server what you want... it just sends you everything. If somebody ever wants to customise it, then maybe that's best left as a server config parameter and not a project config parameter (e.g. file in `:root-dir` or `~`).

aemoncannon:
@fommil Yeah, that was probably premature optimization on my part.

Issue #183, opened 2015-07-01
Add ability to run tests from difference source sets


jacobono:
I was almost going to be pressured to switch to in IDE (after the team heard I use Emacs) and I was so happy to be able to smile a big smile and say "I can do what you can do in Emacs with ensime".  :smile: 

So it goes without saying, that I love this package!

I came across a particular use case on a project with an integration test source set, and unfortunately I can't run any `ensime-*-test` commands in that source set.

I think it would be cool if it were possible to do this -- and I'd be more than happy to work on it!

I was thinking that maybe in this case we could incorporate [hydra](https://github.com/abo-abo/hydra) as it seems fitting to me.

Maybe create a custom variable with a default hydra 

- test
- it:test

and some of its other variants, to feed those in to `sbt-command`.  Then if other users need some more additions, they can define their own.

The hydra was just an idea I thought might be cool -- I don't have much else other than the idea, but I would be happy to work on it if you liked the gist of it.

Sorry if this has been implemented and I didn't see it, or if their was another issue open like this.

fommil:
ha! Funny, you know I discovered this limitation while creating the Scala Days talk https://t.co/NYOc5CDy7K

I think it would be easy to fix, actually, so it would be awesome if you could have a look. It boils down to regexes I think. https://github.com/ensime/ensime-emacs/blob/master/ensime-goto-testfile.el#L241

jacobono:
I had some time to take a look and adding another string match 
```elisp
(string-match-p "src/it/scala/$" dir)
```
helps find an integration test in the `it` directory.

Just to check, I added the same test name for an implementation file in `test` and `it`.
Is this a valid use case? I've seen it happen.

Anyway, currently only picks one.

What do you think of [projectile](https://github.com/bbatsov/projectile)? 

I think it is pretty awesome -- I did a little bit of hacking to try and get a different answer for the (possibly) invalid use case I mentioned above, and I was able to get it to work with some minor changes and show a nice choice option in the `mini-buffer` of which test to select.

I thought that was cool.

It looks through all the file names in the project and see if they match against a Suffix `["Spec" "Specification" "Test" "Prop"] and gets the matching file names.

I guess the downside is it only works on file names, not from the top level class.

Could also change functionality in `ensime-get-test-class-info` (and what it feeds in to) to work on lists, but I thought it might be neat to bring in projectile since it is so awesome (top 10 downloaded packages in MELPA or something like that).

What do you think?

fommil:
[...]

PR #192, opened 2015-07-11
Find test file in other source sets

jacobono:
Integration tests in `src/it`
Functional tests in `src/fun`

Part of #183


fommil:
thanks! this can be easily merged when the tests go green.

fommil:
we've got a lot of transients in the emacs tests at the moment, it just hangs, so we need to keep hitting build again when that happens. We really need to fix it as a priority.

PR #195, opened 2015-07-13
dwim run tests from different source sets

jacobono:
Other part of #183.

There is going to be a bit of clean up, I just wanted to get your thoughts on this addition.

If I'm in the `it` directory and want to run a test, I think it should just run integration tests -- same for functional.  All others default to running the command in unit tests.

You could run this in a non file related buffer (like `magit` or `dired`) and for now I thought you could default to run unit tests.

What I would really think would be cool, is if the catch all case in `cond` or the case of running from a `magit` or `dired` buffer is hit -- a hydra pops up in the mini buffer to help  decide which source set to run the test command in.


fommil:
Cool. I'll look in more detail later. One thing else that would be great: restrict to the current SBT Project. E.g. if I'm hacking in ENSIME's "server" module and want to run the tests for the current file, then prepend "server/" to the call.

BTW, if you want to really improve ScalaTest support look in the server issue tracker for the ticket on that topic. I'm sure bill venners would be delighted to help.

fommil:
this seems to share code with your last PR, can you put the test directory detectors in a common place?

Also, I noticed that your last change now prefers `it` then `fun` then `test`... hmm, I'm not entirely sure about this. How about prompting the user when multiple options are available?

jacobono:
I definitely will move the shared code.  I just wanted to get your input before I went too far.
This was just a rough cut to show the idea of what I was going for.

In the `dwim` case, I don't think the order of the `cond` is important, as those are non-intersecting source sets.  i.e. if the `file-name` is in the `it` source set it doesn't matter what order that is in the `cond`, because it is the only condition that is going to be hit.

Maybe I'm wrong here? It isn't a problem to move the order though.

I would like to prompt for option if the user isn't in one of those directories already.  Otherwise a user would call the function expecting the test to be run under that source set.

I would like another set of functions that aren't do-what-i-mean that prompt with a list of options as well.  And the cases where you can't resolve a source set, would just run a prompt that goes like

- Pick test command to run (`test` or `testQuick`)
- Pick which source set to run it in (`unit` -- default, `it` -- integration, `fun` -- functional)

I didn't want to do all that until I got your buy in, though :smile:

jacobono:
I added a hydra to give user a prompt.

ensime-hydra-sbt

This happens when `dwim` can't resolve which source set to run in, and shows the prompt with the ensime project.

Can also just call this interactively or with a binding.

fommil:
you'll need to add hydra to the ELPA dependency list

```
Debugger entered--Lisp error: (file-error "Cannot open load file" "No such file or directory" "hydra")
  require(hydra)
```

jacobono:
Double checking, is that the list in `ensime.el`?

so add the hydra dependency there and I can remove it from `ensime-sbt.el`?

fommil:
it's in the `-pkg.el` file

jacobono:
thanks for  pointing me to that file!

fommil:
Cool! Can you think of a way to add a unit test to this? Otherwise, plane squash your commits and I'll try it out tomorrow.

fommil:
Is there any way you could lookup which module the test belongs to and prepend that to the SBT command? We know this from the .ensime file.

jacobono:
I rebased off `upstream/master` and squashed commits.

jacobono:
> Can you think of a way to add a unit test to this?

I could write some `ert` tests.

> Is there any way you could lookup which module the test belongs to and prepend that to the SBT command?

Is this for like sub projects?  In my `.ensime` I only see 1 project -- `root`.

fommil:
We already have tests in the ensime-tests file, if you could add some coverage for these new features, that would be great!

Re: multi module, yes. Try it out on the ensime master code and see what I mean.

Re hydra, wow that's a really cool library! If never seen it before. I might have the play with it tomorrow, especially window pane alignment. But it's this maybe a bit heavy weight for our use? All we need is to choose one option from a list so maybe a standard interactive choice might be better. Does that make sense?

jacobono:
Maybe I might need some pointers with testing.  I don't see an 'ensime-tests' file.  I see a `test` folder and one `el` file that looks like it is just loading things up?

I'll check out the multi module stuff.

If you think `hydra` is too much, I can remove -- I just thought it was nice to have choices laid out, and be able to call them with `ft` or `fk` from one menu.

fommil:
Re tests look at `ensime-tests.el`.

Re hydra actually I think you can get that with other approaches (possibly built in to Emacs). Hydra seems to be if you wanted to be in a mini mode to use multiple one-key choices.

BTW thanks for looking into this. If we gave you a small bounty as a thankyou would you claim it when done?

jacobono:
Thank you for offering :smile: Don't worry about the bounty, I'm just happy to contribute -- I use ensime every day of the work week.

Besides, I'm the one bugging you for help!

I'll start working on the other things we have talked about.  Hopefully soon!

fommil:
sweet, thanks! (Interestingly, almost nobody actually wants the bounties... money and free software just doesn't seem to mix. Honestly feel free to take some for a few beers on us :smile:!)

jacobono:
Tests are running from unit test and integration test source set.

I didn't know how to really set up a functional source set in sbt -- so I didn't test that.

Starting to look at finding the subprojects and prepending them to the sbt command.
Can you point me in the right direction to start looking how to query the config file?

fommil:
@jacobono get hold of the `.ensime` config file and find out the `name` of the `subproject` that the file is under. That `name` is the value that you'll need to prepend.

BTW, `fun`ctional tests are not actually supported by the ensime sbt generator so you'd end up having to hack files by hand anyway. I have one at a project in work and I manually add the source and target dirs.

jacobono:
@fommil thanks for the pointer.

prepending of the subproject is done :smile: 

Last thing to check off is the user prompt -- do you have any ideas for that?

Might be kind of cool for a prompt that asks:
- which subproject to run in
- which source set to run in
- which test command to run

jacobono:
I think I know why the tests might be failing.

For the integration tests, I couldn't get into the plist for subprojects and overwrite the default value -- so I ran `gen-ensime` during the test to get it right.

Drone probably doesn't have the plugin to issue that command i'm guessing.

But it was indeed working locally with the subproject prepended.

Will have to look at more later.

fommil:
Nice! I'll look later. I think standard interactive prompting in what you need. And if there is only one module there is no need for a module name of course.  What do you mean by source set and command?

jacobono:
The prompt would be like
- what subproject (module) do you want to run from (`root`, `core`, etc.)
- ok, what source set should we run from (`unit`, `it`, `fun`)
- what command should we run (`test`, `testQuick`)

If there is only one module (subproject in `config` file) you can still prepend the module name, because it won't matter right?  I tested it on my 1 module projects and I can still use `root/it:test` or `root/testQuick`.

^^ This way I didn't have to check if there was only one module or not.

fommil:
I think sbt might still prefer to not have the name in single module projects, I'm never too sure what magic goes on.

And the module / source set gets filled in automatically if invoked from a file or directory that can be resolved, yeah?

It might be an idea to have a default value for the command, `defvar` as `nil` and if that is non-nill then always use it instead of asking the user.

I'd actually want the result of this to populate a minibuffer that I can then press ENTER on or customise further.

One day we might want to extend this with glob pattern restrictions to the spec and/or individual test.

fommil:
ping

ensime-sbt.el
```diff
@@ -88,26 +89,82 @@
...
-                (return (message "Could not find top-level class"))))
+	  (or (ensime-top-level-class-closest-to-point)
+	      (return (message "Could not find top-level class"))))
 	 (cleaned-class (replace-regexp-in-string "<empty>\\." "" impl-class))
 	 (command (concat "test-only" " " cleaned-class)))
-    (sbt-command command)))
+    (ensime-sbt-test-dwim command)))
+
+(defun ensime-guess-project-dir ()
+  (let ((ensime-file (ensime-config-find-file
```
fommil: I've been doing some hacking in this area... there is a much better way to do this.

```elisp
  (let* ((connection (ensime-connection))
         (config (ensime-config connection))
      ...)
```

fommil:
Ping

fommil:
@jacobono hah, I just realised you wrote the grade mode for emacs. I'm going to be using it extensively at my new contract, where gradle is the build tool. Hopefully we can give gradle first class support like the `sbt` integrations.

fommil:
@jacobono ! ping

jacobono:
Sorry this took me so long!

Had quite a few things change this past month, and it took a little while to get acclimated to them.

I finally added the prompt, just tested everything out on `ensime-server` and it works nicely.

Thanks for the tip getting the `config` it was much nicer.

I rebased, so hopefully things still work (shouldn't take long to fix if they don't) -- probably need to change some keybindings later.

fommil:
My turn to be slow, I'm on holiday for another 3 weeks. I will definitely look when I come back unless someone gets there first.

jacobono:
ha its all good. have a great holiday!

ensime-test.el
```diff
@@ -2073,26 +2087,138 @@
...
-    ((:inf-repl-exit)
-     (let ((proc (ensime-test-var-get :repl-proc)))
-       (ensime-assert-equal (process-status proc) 'exit)
-       (ensime-assert-equal (process-exit-status proc) 0)
-       (ensime-test-with-proj (proj src-files) (ensime-cleanup-tmp-project proj)))))
-  ))
-
+(ensime-test-cleanup proj))))
+
+(ensime-async-test
```
fommil: wow, you wrote tests!

fommil:
this is a great PR, thank you! We might one day depend on hydra, in which case we can revisit the implementation.

fommil:
squashing, but history should be retained in this PR. see #274

ensime-test.el
jacobono: haha! :wink:

jacobono:
Thanks for all your help! :+1:

PR #199, opened 2015-07-21
enable server logging during testing

fommil:
I can't work out what is causing the transient failures, but this increases the ability to debug any failures in the emacs tests so is useful in its own right.


aemoncannon:
Nice

Issue #190, opened 2015-07-09
transient test failures


fommil:
We're getting a lot of transient failures these days, on pretty much all the async tests (which implies there is one thing that's systematically going wrong). I'm seeing this locally as well as on drone.

This is the latest example https://github.com/ensime/ensime-server/pull/1038

```
Starting test: 'Test implicit notes.'
Copying /tmp/ensime_test_proj_2228CNm/cache to /tmp/ensime_test_proj_2228QYz/cache
ENSIME server starting...
Received test event: :return-value
Got :return-value, expecting (:connected :compiler-ready :full-typecheck-finished). Ignoring event.
Connecting to Swank on port 56249..
Received test event: :full-typecheck-finished
Received :full-typecheck-finished expecting (:connected :compiler-ready :full-typecheck-finished). Waiting for the rest
ENSIME ready. Hacks and glory await!
Waiting for connection-info
Received test event: :indexer-ready
Got :indexer-ready, expecting (:connected :compiler-ready). Ignoring event.
Received test event: :connected
Received :connected expecting (:connected :compiler-ready). Waiting for the rest
Connected to ENSIME speaking protocol 0.8.17, please wait while the project is loaded.
Received test event: :return-value
Got :return-value, expecting (:compiler-ready). Ignoring event.
Received test event: :compiler-ready
Handling test event: :compiler-ready/(:events (:compiler-ready) :func (lambda (G11202) (ensime-test-run-with-handlers Test implicit notes. (ensime-test-with-proj (proj src-files) (goto-char (ensime-test-before-label 1)) (ensime-assert-equal (ensime-implicit-notes-at (point)) (quote (Implicit parameters added to call of stringToB("xxx"): (zz: scala.Int) Implicit conversion of "xxx" using stringToB: (s: String)(implicit x: Int)pack.B))) (ensime-test-cleanup proj)))) :is-last t)
Received test event: :region-sem-highlighted
Received test event: :return-value
```

fommil:
another when building #192 

```
Starting test: 'Test file with package name (this broke when -sourcepath param was used).'
Copying /tmp/ensime_test_proj_44yaC/cache to /tmp/ensime_test_proj_44mKP/cache
ENSIME server starting...
Connecting to Swank on port 50714..
Received test event: :indexer-ready
Got :indexer-ready, expecting (:connected :compiler-ready :full-typecheck-finished). Ignoring event.
Initializ
Received test event: :connected
[...]

PR #200, opened 2015-07-23
customise logback properly

fommil:


fommil:
I'm going to self merge this when it goes green because I think it might have found #190

PR #202, opened 2015-07-24
get undercover working again

fommil:


fommil:
damn, hit a transient. It would be good if drone allowed saving some log files matching a glob pattern.

fommil:
hmm, two transients in a row... I'm starting to think undercover is highlighting a problem as a side effect.

PR #208, opened 2015-08-08
Path urls

fommil:
@aemoncannon I asked on the sublime and atom channels and neither of them have hooked up to the docserver yet, so I reckon it'll be a lot easier to make the change to emacs than a massive hack in the server to include the hostname.

Given that both sublime and atom are currently in the process of moving over to the websockets-only protocol, it'll be really simple for them to handle relative paths. This seems a bit heavyweight for emacs right now, but it's step one in moving to SWANKY (natural S-Express form over WebSockets).


fommil:
err, I broke all the things.

aemoncannon:
Okey dokey.

fommil:
pushed a fix... this should be good. @aemoncannon @edani I just discovered Cask and its awesomeness. With this small change, you can set up flycheck to have all the ENSIME dependencies in its linter! That means the error messages it complains about are actually sensible! Read the `Cask` file for details.

I've asked if MELPA can automatically generate the pkg file for us, so we don't have to worry about it.
 https://github.com/milkypostman/melpa/issues/3001

fommil:
this is still broken, but tests aren't catching it... subtle thing to do with the way `concat` works, fix soon.

fommil:
ok, I fell dumb. Was using `append` instead of `concat` doh! This is working for me.

fommil:
since you've already had a look at the main changes in this, I'm self-merging so that the server side change can go green.

PR #209, opened 2015-08-08
preemptive support for new server features

fommil:
and this has the keybindings for the new server features. I'll wait until the server support is merged before updating the wiki.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,7 +1,8 @@
 image: ensime/ensime:latest
 
 script:
-  - test/run_emacs_tests.sh &&
+  - cask pkg-file &&
+    test/run_emacs_tests.sh &&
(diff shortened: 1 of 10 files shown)


fommil:
meh, I'm not sure how to get this to pass for the current server code. I'll think about it tomorrow.

```
Error executing test: (ensime-test-interrupted . Test interrupted: (ensime-test-assert-failed . Expected truth of (let ((url (ensime--inspector-doc-url-at-point))) (s-ends-with-p java/lang/String.html#equalsIgnoreCase%28java.lang.String%29 (url-hexify-string url))).).), moving to next.
```

fommil:
I'm going to update the docker images to have Cask installed before merging this.

fommil:
because of the regression I introduced, I'm merging this as a priority

PR #235, opened 2015-09-07
Replace custom when-let with dash -when-let

lunaryorn:
The custom `when-let` from `ensime-macros.el` macro is incompatible with `when-let` from `subr-x.el` introduced in Emacs 25, which can cause subtle breakage if a caller expects to get the `subr-x` variant, but ends up with the one from Ensime or vice versa.  See https://github.com/Malabarba/paradox/issues/95 for an example.

This PR replaces the custom macro with `-when-let` from dash.el, which comes at no additional cost since Ensime already depends upon dash.el anyway–I wonder why `when-let` was defined anyway.

Besides, a package should _never_ define unprefixed symbols, so if `when-let` needs to stay for a reason that I'm not aware of, it needs to be renamed (i.e. `ensime-when-let`).

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -44,6 +44,7 @@
   (require 'cl)
   (require 'ensime-macros))
 (require 'ensime-vars)
+(require 'dash)
 
 (defvar ensime-net-processes nil
   "List of processes (sockets) connected to Lisps.")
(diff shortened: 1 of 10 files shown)


fommil:
Sweet. ENSIME may predate dash, so there's a load of stuff like this that can be retrofitted.

lunaryorn:
@fommil Will you merge it?

fommil:
I was waiting for the CI to finish.

fommil:
Thanks!

fommil:
Oh cool, you're the flycheck dude. How hard would it be the move our warnings / errors to flycheck?

lunaryorn:
@fommil I don't know.  Generally it's possible, and I have [experimental Flycheck support for Ensime](https://github.com/lunaryorn/.emacs.d/blob/master/lisp/flycheck-ensime.el) in my Emacs configuration, but last time I tried it it would occasionally look up Emacs.  I haven't had time for further debugging since then.

It's probably a sign that I need to open up Flycheck's API for this purpose, but that's a larger story which I don't have time for currently. 

Feel free to take it and try it.

fommil:
we're happy to help with the debugging on the ensime side if you ever get time to look at this again.

lunaryorn:
Can't promise anything, Ensime's own reporting is too good, and you know, the good is the enemy of the better :)

Besides, as far as I can see Ensime now shows notes for implicit conversions at point, and that's not easily ported to Flycheck.

fommil:
interesting. Doesn't flycheck have support for showing information on a line?

lunaryorn:
Well, it _can_ show annotations on a specific line, but it has no API to _dynamically create_ the annotations when the user requests all annotations at a specific positions.  Flycheck wants all annotations upfront.

fommil:
hmm, interesting and noted. We might be able to do that, I don't think it's particularly expensive.

PR #238, opened 2015-09-14
Assembly

fommil:
this allows a user-provided `-assembly.jar` file of the server to trump the `sbt` update stage. Which has the fortunate side-effect of simplifying the dev cycle.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -11,5 +11,5 @@ build:
 
 matrix:
   EMACS:
-    - /opt/emacs-24.4/bin/emacs
+    - /opt/emacs-24.5/bin/emacs
     - /opt/emacs-24.3/bin/emacs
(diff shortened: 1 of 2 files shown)


aemoncannon:
Neat! For dev, what advantage does this have over 'sbt publish-local'?
Boots faster?

fommil:
@aemoncannon stands a chance of me being able to, err, get it into a certain corporate environment that I've recently become acquainted with (whistle whistle whistle)

fommil:
@aemoncannon oh, but actually it is better for dev... because there is less reliance on sbt and ivy actually getting `publishLocal` right.

fommil:
btw, I also updated the docs here https://github.com/ensime/ensime-emacs/wiki/Quick-Start-Guide#installing-the-server-from-source

fommil:
(I have a corresponding PR into the server)

PR #250, opened 2015-09-27
Don't send full source file contents over socket.

aemoncannon:
The sexp protocol parsing layer has trouble with big files. For example a 1000 line java file might take 1+ seconds to parse. This change falls back on an alternative scheme where the source contents are shared via a temporary file. See ensime/ensime-server#1102

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -997,10 +997,9 @@ copies. All other objects are used unchanged. List must not contain cycles."
 (defun ensime-rpc-async-typecheck-files (file-names continue)
   (ensime-eval-async `(swank:typecheck-files ,file-names) continue))
 
-(defun ensime-rpc-async-typecheck-file-with-contents (file-name contents continue)
-  (if (version< (ensime-protocol-version) "0.8.11")
-      (ensime-eval-async `(swank:typecheck-file ,file-name ,contents) continue)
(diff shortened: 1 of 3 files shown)


fommil:
you'll need to rebase to upstream as the drone file format changed

ensime-client.el
```diff
@@ -996,10 +996,9 @@ copies. All other objects are used unchanged. List must not contain cycles."
 (defun ensime-rpc-async-typecheck-files (file-names continue)
   (ensime-eval-async `(swank:typecheck-files ,file-names) continue))
 
-(defun ensime-rpc-async-typecheck-file-with-contents (file-name contents continue)
-  (if (version< (ensime-protocol-version) "0.8.11")
```
fommil: we could probably remove a tonne more of these checks because everybody should be using the latest version of ensime-server now.

aemoncannon:
Ok. Also going to make some changes to the way the temp file is named.

aemoncannon:
PTAL - I'm allowing for 25 non-interfering file channels right now...

fommil:
BTW, at my work the classpath file and/or fat jar will be on a read-only network drive so I was probably going to file with the definitions in a few weeks anyway.

aemoncannon:
FYI drone gave false green when I had a syntax bug in ensime-client.el

fommil:
Damn. I think I've seen a few of those. We might just have to be extra careful for a few weeks.

Issue #253, opened 2015-09-29
localise all the sbt-related functions, including the starter


fommil:
not everybody has the luxury of an `sbt` install on their machine (typical for corporate environments), therefore `sbt` and `sbt-mode` should be considered optional dependencies.

It would be good to move all sbt related functions into the optional `ensime-sbt.el`, including the sbt specific parts of the server starter (which should be redesigned to simply require the provision of a class of classpath entries, with various implementations).
##

PR #254, opened 2015-09-29
remove hard auto-complete dependency

fommil:
This isn't the PR that I really wanted to submit in this area, because I think there is a lot more can be done on separating out the sbt dependency... so I left it as a required (not optional) dependency and raised #253 

The motivation for this PR is that I want to reduce the dependencies to make it easier to bring ENSIME into corporate environments (most notably, **my** corporate environment).

Making `auto-complete` optional is a bit of a hack... it seems that neither `cask` nor emacs package really understand the concept of an optional dependency. So here's the hack:

* remove `auto-complete` from the dependency list
* add it to the `Cask` file as a developer dependency so that it is available for testing (actually we don't have any tests for it, so this is blank)
* remove all `(require 'ensime-auto-complete)` and replace with autoloads for the useful function
* remove all `(require 'auto-complete)` and replace them with autoloads for the functions that we need, allowing `package` to compile the files on desktops that **don't** have `auto-complete` installed

I tested this on a pristine Emacs install with and without `auto-complete` installed. Without, `ensime` compiled and the user gets an error if they select `auto-complete` completion style but don't have it installed. With, everything works as expected.

After discussing on the `#emacs` IRC channel, people tend to publish multiple packages to deal with optional components. That seems a bit heavy, so maybe this is a hack we're happy enough with.

Oh, one new testing feature: we now use cask to compile the lisp files (which serves as a lint stage that we're ignoring) and to download all our dependencies from MELPA stable.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -4,6 +4,8 @@ build:
     - SCALA_VERSION=2.11.7
   commands:
     - cask pkg-file &&
+      cask install &&
+      cask build &&
       test/run_emacs_tests.sh &&
       if [ -n "$UNDERCOVER" ] && [ -f "coveralls.json" ] ; then
         curl -v -include --form json_file=@coveralls.json https://coveralls.io/api/v1/jobs || true ;
--- a/Cask
+++ b/Cask
@@ -22,8 +22,15 @@
 ;;
 ;;; Code:
 
-(source melpa)
+(source melpa-stable)
 
 (package-file "ensime.el")
+
+(development
+ ;; optional dependencies
+ ;;(depends-on "auto-complete")
+
+ ; genuine dev dependencies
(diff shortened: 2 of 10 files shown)


edani:
@fommil  This look good to me, is it ready to merge?  Also how about updating the doc of `ensime-completion-style` and say that `auto-complete` needs to be installed manually?

fommil:
@edani ready :-) bugger, you're right I meant to update the docs. I can't make any changes for 3 weeks, no laptop. And also no room in my hands unless I put down this massive glass of pina colada :-P

fommil:
I found that we are using `popup` for `ensime-import-type-at-point`, so I'll have to add that back into the list. It was previously a transient dependency of `autocomplete`.

PR #255, opened 2015-10-01
delete duplicate classpath entries

tanzoniteblack:
When using ensime with multiple submodules with overlapping
dependencies, I was finding that the exact same jars were being added
multiple times to the classpath. Java doesn't consider this an issue,
but the classpath was long enough due to this issue to exceed the
maximum size allowed for bash on my machine (argument list too long exception).

This is not actually a proper fix for the args list being too long, but should prevent it from happening quite so often.

### DIFF
--- a/ensime-inf.el
+++ b/ensime-inf.el
@@ -175,11 +175,11 @@ the current project's dependencies. Returns list of form (cmd [arg]*)"
          :java (concat (plist-get config :java-home) "/bin/java")
          :java-flags (or (plist-get config :java-flags) ensime-default-java-flags)
          :classpath (ensime--build-classpath
-                     (apply #'append
-                            (ensime--scan-classpath (ensime-read-from-file (ensime--classpath-file (plist-get config :scala-version)))
-                                                    "\\(scala-compiler\\|scala-reflect\\)-[.[:digit:]]+\\.jar$")
-                            (funcall get-deps config)
-                            (mapcar get-deps (plist-get config :subprojects)))))
+                     (delete-dups (apply #'append
+                                         (ensime--scan-classpath (ensime-read-from-file (ensime--classpath-file (plist-get config :scala-version)))
+                                                                 "\\(scala-compiler\\|scala-reflect\\)-[.[:digit:]]+\\.jar$")
+                                         (funcall get-deps config)
+                                         (mapcar get-deps (plist-get config :subprojects))))))
       (error "No ensime config available"))))
 
 (defun ensime-inf-switch ()


fommil:
I'm not sure this does what you think it does. This is the classpath of ENSIME itself, not your project.

tanzoniteblack:
My understanding of this chunk of code is that it aggregates the classpath variables mentioned in the `.ensime` file using the `get-deps` function defined in the `let` and combines those with the jars needed for running ENSIME itself. So it's a combination of the classpath of ENSIME itself and my project. However the way it's written allows for duplicates if any of the deps sections in the submodules of the `.ensime` file overlap.

I'm also not clear that the build failing had anything to do with my changes?

fommil:
Hmm, that sounds wrong.

tanzoniteblack:
For reference sake on my particular issue, it knocks the number of jars being thrown into the classpath here from 1889 to 391. And they are definitely not only jars that are being asked for by ensime itself, because they contain private jars to my company.

Also, for reference sake, this bit me when I was running the command `(ensime-inf-switch)` (via `C-c C-v z`) in a project to try something out in the repl.

fommil:
Aah! We scan the classpath when launching the REPL yes. You must have a lot of leaf node projects. I think you'll need to check the failed tests. We can't merge unless everything in green.

tanzoniteblack:
@fommil if you take a look at the build errors, it's not a particular test that's failing, but the build system itself is panic-ing due to a failure. Same failure happened on a forced repush to cause the rebuild. I'll work on seeing if I can replicate this locally or not.

Obviously we shouldn't merge this until the builds succeed though :)

fommil:
Sorry I can really look at this for 3 weeks as I'm on holiday. I'll try to hit the rebuild button for you.

fommil:
Ah crap, looks like drone introduced a regression. We're without a CI unless @rorygraves can log in to fommil.com and install the latest drone 0.4 snapshot. Should just be a wget and a dpkg install (both in the bash history)

fommil:
can you please rebase to upstream? the CI issues should have resolved themselves

tanzoniteblack:
Rebased upstream, and now the tests failed for 24.5 (not 24.4 though), but at least it wasn't a panic before emacs was even loaded. I'll try looking into the tests later when I have some time.

fommil:
hmm, I've restarted that build incase its a weird transient failure

fommil:
btw, install  http://readme.drone.io/docs/cli/index.html and type `drone exec` if you want to run the CI locally

fommil:
it failed to restart (not sure why)

tanzoniteblack:
Force re pushed the collapsed changes on master and the build passed this time.

fommil:
OK I'll have a look

fommil:
thanks!

PR #256, opened 2015-10-09
Performance fix when typechecking

edani:
`ensime-config-includes-source-file` is pretty expensive, and was called before every automatic typecheck. For projects with large configs such as ensime itself, I was seeing emacs freeze for 0.5s .

The fix here is to minimize calls to `ensime-config-includes-source-file` by assuming that a file can belong to only one ensime connection. I think that's a pretty reasonable assumption. Let me know if you disagree and I'll think of some other solution.

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -141,9 +141,7 @@ overrides `ensime-buffer-connection'.")
   "Return the connection to use for ENSIME interaction in the current buffer.
  Return nil if there's no connection.
    * In most code we prefer (ensime-connection), which raises an error if
-     the connection is not present.
-   * This function is ambiguous if there's more than one ensime connection for
-     the current source file (shouldn't really happen in practice)."
+     the connection is not present."
   (or (ensime-conn-if-alive ensime-dispatching-connection)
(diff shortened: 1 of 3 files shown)


aemoncannon:
Good catch! I agree two connections managing the same source file is a pretty weird case.

aemoncannon:
LGTM

fommil:
Great! I didn't quite realise it had such a hit to performance. I'd like the revisit/share a bunch of this code with the swanky websockets stuff I plan to do in November.

fommil:
I'm finding myself in the situation where I need to find the server process for the current file (or directory) and it's actually quite tricky. One thing we probably need is the ability to handle directories from `dired` as well as source scala/java files, which might be the root dir itself.

aemoncannon:
FYI I think my emacs PR overlaps with this a bit. I added caching to
ensime-config-includes-source-file to make it scale to hundreds of source
roots. I'll rebase tonight.

fommil:
I have also made a few changes to the "find process for source file" function that you might want to check as well but they are really minor.

PR #262, opened 2015-10-19
Misc. fixes to make ensime play better with a big, pure java project.

aemoncannon:
* Add a temporary adaptor function to make param expansion work for
  java. Completion is *nice* in java now :-)

* Document (with error) that java home is required in config. This was always the case, but we died with a crap message.

* Don't try to fan out typecheck requests to multiple connections. This
  triggers expensive connection-for-sourcefile lookups that choke when
  there are *many* source roots.

* Guard our find-file and save-file hooks with ensime-source-file-p. We
  were previously taxing *all* buffers with expensive connection
  lookups.

* Handle nil semantic highlighting (a hack really).

### DIFF
--- a/ensime-config.el
+++ b/ensime-config.el
@@ -60,6 +60,12 @@
       (error "Name in ensime configuration file appears to be unset"))
     name))
 
+(defun ensime--get-java-home (config)
+  (let ((value (plist-get config :java-home)))
+    (unless value
+      (error "java-home in ensime configuration file appears to be unset"))
+    value))
+
 (defun ensime-config-source-roots (conf)
   "Returns a list of all directories mentioned in :source-roots directives."
   (let ((subs (plist-get conf :subprojects)))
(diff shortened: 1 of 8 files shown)


fommil:
Awesome! I am back on Friday and I have a free pass from The Boss to work on ENSIME for three days, I'll look at all your Java stuff for definite and send a PR to you with any suggestions.

aemoncannon:
Still needs scalability fixes for source-roots before this is ready. My plan is to sort the roots and use binary search to determine if the file in our roots. Should be muuuuch faster.

rorygraves:
:+1:  Awesome stuff!

fommil:
Interesting, I never realised that was a performance bottleneck. I've used ENSIME on mostly Scala projects and I guess there were only about 50 source roots Max. I'd love to change the .ensime format one day to account for different phases (main, test, it, fun, etc) but that's ages away. It would allow us to spawn up separate PCs for different files and have the proper classpath for every file.

fommil:
this has introduced a dependency on https://github.com/skeeto/emacs-memoize do we really need it?

aemoncannon:
need it? not at all, but it streamlined that one section of code and is a
handy tool. is it disruptive to add new elisp deps, or are you just
concerned about code size?

fommil:
it would be good if we keep the dependencies at a minimum for a couple months, while I onboard ENSIME at my work (every package gets vetted and can slow everything down). Oh god, I've just realised that the inclusion of netbeans in the server is going to kill any chance of me getting the non-assembly jars accepted.

fommil:
actually, it might not be that bad. I could find the URLs for all the jars and then just submit them as one package called "netbeans-lib" or something like that.

fommil:
I was hoping to tidy up https://github.com/ensime/ensime-emacs/pull/254 and get it merged pretty soon to help with this effort.

fommil:
I'm getting `let*: Symbol's function definition is void: popup-menu*` when doing `inspect-type-at-point`

aemoncannon:
Oh. We don't support that yet. You're hitting one of my really old java
code paths.

aemoncannon:
(try print-type-at-point)

fommil:
I'm getting hangs with "Illegal State: Compilation in Progress".

Gotta go eat now.

fommil:
here's my test project: https://github.com/fommil/matrix-toolkits-java/commit/1b2d229024142a029d05c1f1f5450d9623cbd45e basically check out master, copy/paste that `build.sbt` from the commit comment and then run `sbt gen-ensime`. It's definitely doing something! But my Java 6 "fixes" are not backwards compatible so it breaks on Java 7.

fommil:
I'm getting conflicts now, but I did have this working earlier and it was very impressive! I'll wait for you to rebase before having another look.

The biggest question I have really is "do we use `ensime-mode` for Java as well?" or should we create an `enjime-mode` that adds java specific functions. That would also allow us to simplify the API so that it has language specific requests and responses.

aemoncannon:
Rebased. PTAL

I do think we should move toward enjime-mode, if for no other reason than to give accurate keymaps / menus for Java users. Also fine with a new family of RPCs, so long as we're not gratuitously changing names (I don't want to create extra work for the clients).

ensime-config.el
```diff
@@ -70,20 +76,43 @@
...
-      (-first (lambda (dir) (ensime-path-includes-dir-p file dir))
-	      source-roots))))
+    (let ((dir-set (ensime--source-root-set conf no-ref-sources)))
+      (let ((d (file-name-directory (expand-file-name file))))
+        (catch 'return
+          (while d
+            (let ((prev d))
+              (when (gethash (file-truename d) dir-set)
+                (throw 'return t))
+              (setq d (file-name-directory (directory-file-name d)))
```
fommil: Should we perhaps normalise when we load the config for the first time? Just mutate all the dir references.

PR #263, opened 2015-10-21
Add support for Expand Region

lunaryorn:
See https://github.com/magnars/expand-region.el/pull/181

### DIFF
--- a/ensime-expand-region.el
+++ b/ensime-expand-region.el
@@ -1,0 +1,61 @@
+;;; ensime-expand-region.el --- Ensime expansions for Expand Region  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2015  Sebastian Wiesner
+
+;; Author: Sebastian Wiesner <swiesner@lunaryorn.com>
+
+;; This file is not part of GNU Emacs
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Add Ensime expansions to Expand Region.
+
+;;; Code:
+
+(require 'ensime-mode)
+(require 'ensime-client)
+(require 'ensime-editor)
+
+;; Define Expand Region variable explicitly to avoid a hard `require' on Expand
+;; Region.  Since we only enable this part of Ensime if Expand Region is
+;; actually available, that's pretty safe nonetheless :)
+(defvar er/try-expand-list)
+
+(defun ensime-expand-region-mark-syntactic-context ()
+  "Mark the next outer syntactic context."
+  (when (ensime-connected-p)
+    (ensime-expand-selection (mark) (point))))
+
+(defun ensime-expand-region-add-expansions ()
+  "Add expansions for Ensime."
+  (set (make-local-variable 'er/try-expand-list)
+       (append er/try-expand-list
+               '(ensime-expand-region-mark-syntactic-context))))
+
+(defun ensime-expand-region-enable ()
+  "Enable Ensime expansions for all Ensime Mode buffers."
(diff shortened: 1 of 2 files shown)


fommil:
could you please rebase to upstream master?

ensime-expand-region.el
```diff
@@ -0,0 +1,61 @@
+;;; ensime-expand-region.el --- Ensime expansions for Expand Region  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2015  Sebastian Wiesner
+
+;; Author: Sebastian Wiesner <swiesner@lunaryorn.com>
+
+;; This file is not part of GNU Emacs
```
fommil: I'm ok with the shorter header form of `ensime-http`. This classic form is so verbose...

ensime.el
```diff
@@ -49,6 +49,7 @@
 (require 'ensime-stacktrace)
 (require 'ensime-debug)
 (require 'ensime-editor)
+(require 'ensime-expand-region)
```
fommil: this is adding a hard dependency, it'll fail if `expand-region` is not installed.

Rather add a hard dependency to `(require 'expand-region)` in `ensime-expand-region.el`, and then let users `(require 'ensime-expand-region)` if they want this.

ensime-expand-region.el
```diff
@@ -0,0 +1,61 @@
...
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Add Ensime expansions to Expand Region.
```
fommil: since this will be a standalone file, can you please add instructions of use here?

ensime-expand-region.el
lunaryorn: Shall I change it?  I usually use elide-head to hide this blurb.

ensime.el
lunaryorn: I don't think that this is a hard dependency.  This change does not `(require 'expand-region)` anywhere. 

`ensime-expand-region.el` only `defvar`s a particular variable of Expand Region which is the main extension point, but doesn't use any other code from Expand Region.  And it only becomes active after Expand Region is loaded.

fommil:
I'm ok with this if you're absolutely sure that it won't blow up the compile or people's ensime if they don't have `expand-region` installed.

Can you add some docs though? Especially for people who don't know what `expand-region` is.

lunaryorn:
> I'm ok with this if you're absolutely sure that it won't blow up the compile or people's ensime if they don't have expand-region installed.

I am sure :sunglasses: 

> Can you add some docs though? Especially for people who don't know what expand-region is.

Sure.  Mind to tell me where you have your docs and where I should add documentation for this feature?

fommil:
Just in the header of your file in fine.

fommil:
Actually I was thinking about this again. I'd prefer that we were consistent about the way we deal with optional deps. You don't need to make the change, I can tweak your commit once you're happy with it to be in line with our existing approach (which is going to grow). It'll just mean an extra require in your personal config.

fommil:
Can you think of a sensible unit test for this?

lunaryorn:
@fommil I don't think that there's a reasonable way to test this.  There's no functionality in this code, it's just glue between an existing feature in Ensime and a framework provided by another library.

fommil:
All code is testable, here we could use mocks. But I appreciate that we don't have a standard testing framework in place so it would be a bit cruel to demand it.

lunaryorn:
@fommil I disagree.  If you mock everything away, all that you still test is that you do _something_ with the mocks.  But your tests neither cover functionality nor do they ensure proper integration of all parts.

You'd rather write some kind of integration test for this thing, i.e. with ecukes, but that'd make expand-region a hard dependency of the integration test suite.

fommil:
Yeah, this file is too trivial to necessitate a test. actually that's the approach we take with optional dependencies: we add them to the `Cask` file as test deps. Which means our CI doesn't catch load-path problems with optionals :disappointed:

If you add a few lines describing what this is for, I'll merge. Even better if you follow the same pattern as `ensime-auto-complete.el`.

lunaryorn:
@fommil I'll do that.  Will take me some time, though, I'm pretty busy currently :)

fommil:
see #274 thanks! BTW, you'll need to manually `(require 'ensime-expand-region)` in your config. I just want to be consistent with the way we deal with optional dependencies.

lunaryorn:
@fommil Thank you!

PR #266, opened 2015-10-25
ensime-server will soon depend on netbeans jars

fommil:
@rorygraves @hedefalk @yazgoo this is an update to the `build.sbt` file that we use to download ensime-server and its dependencies. Can you please update your variants of this launcher, as the next PR to be merged into master will need the netbeans repo (and I bet you were probably missing some of the finer details anyway, because I forgot to put them in the reference launch script!)


hedefalk:
Ack

hedefalk:
-= is new?

Ah, I was using 0.13.8. Works with 0.13.9.

PR #268, opened 2015-10-25
Address a couple of niggles that I've been putting off...

fommil:


fommil:
damnit, looks like there is another bug in drone where failed tests are showing up green. We can't trust the green status light until that is fixed, which means manual inspection of the logs... joy.

fommil:
we need this one to get the CI back on track... oops! sorry.

PR #277, opened 2015-11-01
identified a source of transient failures

fommil:


fommil:
nope, that's not it. It looks like the code that detects the connection has somehow broken. I must have merged a PR before it finished with all the commits in it.

fommil:
I think I know what it is. The build has broken for a while because of uses of `&&` vs `;` in the drone file. It's just showing red now, after the PR merges, because I fixed it in trunk and the PRs didn't have the drone "fix" (that breaks the build).

PR #281, opened 2015-11-02
Only require expand region core

lunaryorn:
That's enough to provide the variable definition.  There's no need to load all of Expand Region.


lunaryorn:
@fommil Wow, thanks!  Merging faster than the speed of light :heart_eyes: :+1:

fommil:
(yeah, otherwise you'd see that the CI is bust :stuck_out_tongue: )

PR #282, opened 2015-11-04
Fixes to get elisp CI back up and running.

aemoncannon:
This is probably the first of many necessary fixes.

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -345,24 +345,21 @@ This doesn't mean it will connect right after Ensime is loaded."
          (ensime-dispatching-connection process))
     (ensime-setup-connection process)))
 
-
-(defun ensime-handle-connection-info (connection info)
+(defun ensime-handle-connection-info (info)
   "Initialize CONNECTION with INFO received from Lisp."
-  (ensime-event-sig :connected info)
-  (let ((ensime-dispatching-connection connection))
-    (destructuring-bind (&key pid server-implementation version
-                              &allow-other-keys) info
-      (setf (ensime-pid) pid)
-      (setf (ensime-protocol-version) version)
-      (destructuring-bind (&key name) server-implementation
-        (setf (ensime-server-implementation-name) name
-              (ensime-connection-name) (ensime-generate-connection-name name)))
-      ))
 
-  (run-hooks 'ensime-connected-hook)
+  (assert (ensime-connection-or-nil))
+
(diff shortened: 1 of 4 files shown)


aemoncannon:
Arghhh so many things broken. How long has this been busted? Months?

fommil:
Eek, yeah possibly. I guess I should go learn bash again :blush:

fommil:
BTW, the crazy sit for call is to avoid a problem in subsequent calls which expect the protocol version to be set. I figured the only way to really fix it would be to do that async but it's way beyond my Emacs foo.

Maybe we can use this as an opportunity to remove all the version checks and just assume everyone is on a recent version. They are only there as a temporary migration horror anyway.

fommil:
Or, we could just set the initial default protocol version and remove the sit for. Enough time has lapsed the make it sensible.

aemoncannon:
Yeah. I cleared up a few breakages. I'ma work on this more later.

aemoncannon:
Related: the ensime-server build is failing with timeouts on BasicWorkFlowTest and DebugTest -- why do those tests take so long?

fommil:
They are full integration tests against the example projects, but they shouldn't be timing out on reasonably powerful machines. Is this when you run them locally?

aemoncannon:
Na the github build broke when I merged a revert last night

fommil:
Oh, it broke because that change will fail the tests. It breaks the message expectations of the test. Look it ProjectFixture for where we assert what we expect to receive.

BTW, if you run emacs tests locally you can just `publishLocal` the server and the Emacs tests should use that (once you clean out the temp dir). Or if you want to use drone to do it, you can just change the URL/branch of the emacs that we're testing against (once you uncomment that test).

fommil:
oh, one other thing, it turns out that it won't be too much trouble to add a dependency on the `memoize` package after all. I still have an opportunity to request additional packages over the coming months.

aemoncannon:
Ok. We're back in Greensville. Situation was not as bad as I suggested; most of the breakage was caused by my recent server changes, and the remainder was simple elisp fixes. Will merge when drone is done.

aemoncannon:
Once this is in I'm going to revisit the server events. I think the Broadcast(Persist(...)) doesn't quite do what we want...

fommil:
Great stuff!

PR #284, opened 2015-11-05
Require connection-info first in integration suite.

aemoncannon:
I want to lock this down for posterity.


fommil:
Good idea. We should do the same in the server's tests.

Issue #283, opened 2015-11-05
ensime keeps restarting if a file is not found


oscar-broman:
I created a buffer called `Test.scala` and ensime kept restarting until I saved it.

```
java.io.FileNotFoundException: source file '/Users/oscar/project/src/main/scala/package/Test.scala' could not be found
	at scala.tools.nsc.Global.getSourceFile(Global.scala:372) ~[scala-compiler-2.11.7.jar:na]
	at org.ensime.core.RichCompilerControl$class.createSourceFile(RichPresentationCompiler.scala:177) ~[core_2.11.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.RichPresentationCompiler.createSourceFile(RichPresentationCompiler.scala:198) ~[core_2.11.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer.createSourceFile(Analyzer.scala:290) ~[core_2.11.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer$$anonfun$allTheThings$1.applyOrElse(Analyzer.scala:236) ~[core_2.11.jar:0.9.10-SNAPSHOT]
	at org.ensime.core.Analyzer$$anonfun$allTheThings$1.applyOrElse(Analyzer.scala:152) ~[core_2.11.jar:0.9.10-SNAPSHOT]
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36) ~[scala-library-2.11.7.jar:na]
	at org.ensime.core.Analyzer$$anonfun$ready$1.applyOrElse(Analyzer.scala:149) ~[core_2.11.jar:0.9.10-SNAPSHOT]
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36) ~[scala-library-2.11.7.jar:na]
	at akka.event.LoggingReceive.apply(LoggingReceive.scala:62) ~[akka-actor_2.11-2.3.14.jar:na]
	at akka.event.LoggingReceive.apply(LoggingReceive.scala:50) ~[akka-actor_2.11-2.3.14.jar:na]
	at scala.PartialFunction$class.applyOrElse(PartialFunction.scala:123) ~[scala-library-2.11.7.jar:na]
	at akka.event.LoggingReceive.applyOrElse(LoggingReceive.scala:50) ~[akka-actor_2.11-2.3.14.jar:na]
[...]

PR #286, opened 2015-11-08
Don't attempt to highlight missing files.

aemoncannon:
Temporary fix for #283. The proper fix will accept SourceFileInfo on the server, so we don't need to require a disk file.


fommil:
sorry, I forgot to merge this last night.

PR #288, opened 2015-11-09
Changed the implicit conversion and implicit parameter highlight

noelmarkham:
Changed the implicit conversion and implicit parameter highlight to be a little less in-your-face: light gray straight underline.


fommil:
I call this the John Major.

aemoncannon:
Nice

Issue #247, opened 2015-09-23
Do we really need a gutter icon *and* an underline for implicits?


aemoncannon:
Seems like too much noise to me. Also, can't see breakpoints when gutter is in use.

fommil:
Somebody requested this currently and I recommended that they set the overlay to nil. But in general I wonder if there is a prettier way of doing it. Maybe gutter for implicit params and something else for multiline implicits

metatron-the-chronicler:
Maybe something like Eldoc might work for this.  There could be a setting that says that whenever Eldoc gets activated it should display the implicits in the signature.  It might be preferable to just have the gutter icon for everything in the editor window and a command that shows all of the implicits in that are in the file.

fommil:
This is now fixed actually, but requires a setting.

PR #293, opened 2015-11-15
Fix for #247 - remove the noisy gutter icon for implicits

noelmarkham:
Remove the noisy gutter icon for implicit parameters and conversions.

They're underlined anyway.

### DIFF
--- a/ensime-semantic-highlight.el
+++ b/ensime-semantic-highlight.el
@@ -41,8 +41,9 @@
               (setq start end)
               (setq end (1+ end)))
 	    (let ((ov (make-overlay start end buf)))
-              (when (or (eq type 'implicitParams)
-                        (eq type 'implicitConversion))
+	      (when (and ensime-implicit-gutter-icons
+			 (or (eq type 'implicitParams)
+			     (eq type 'implicitConversion)))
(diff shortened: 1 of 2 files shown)


fommil:
interesting, the 24.3 test failed and it looks like a new version of `epl` introduced an emacs version dependency.

Hmm, I'll restart, but I'm happy for you to change the `.drone.yml` file to do a cross build to 24.4 and 24.5 instead. Nobody uses 24.3 anymore because magit is 24.4+ only.

noelmarkham:
I've updated this to depend on a variable, set to display the gutter icons by default. I've never written elisp before, so please pull it apart if I'm Doing It Wrong.

Also, as suggested, set Drone to run against 24.4 instead of 24.3.

fommil:
Cool but you don't need symbol-value, or quoting the defcustom symbol. Just putting the variable in raw works. nil is false and non-nil is true in LISP.

fommil:
great! sorry to be an arse about this, but could you please revert the 24.3 change and squash everything? 24.3 is green again, so it was probably just a transient.

noelmarkham:
Done :)

PR #301, opened 2015-11-20
Allow scala-compiler version omission in classpath

sirius94:
This patch changes the regexp for matching scala-compiler and scala-reflect jars in the classpath so that it also matches jars without a version number in their names.

### DIFF
--- a/ensime-inf.el
+++ b/ensime-inf.el
@@ -185,7 +185,7 @@ the current project's dependencies. Returns list of form (cmd [arg]*)"
          :classpath (ensime--build-classpath
                      (delete-dups (apply #'append
                                          (ensime--scan-classpath (ensime-read-from-file (ensime--classpath-file (plist-get config :scala-version)))
-                                                                 "\\(scala-compiler\\|scala-reflect\\)-[.[:digit:]]+\\.jar$")
+                                                                 "\\(scala-compiler\\|scala-reflect\\)\\(-[.[:digit:]]+\\)?\\.jar$")
                                          (funcall get-deps config)
                                          (mapcar get-deps (plist-get config :subprojects))))))
       (error "No ensime config available"))))


fommil:
Thank you for using ENSIME!

We are getting a lot of bug reports recently that are missing key information, so we're using https://github.com/Astrac/github-autoresponder to ask that everything is captured in the ticket.

Before somebody responds to this ticket, can you please confirm that you have checked the following:

* are you using the latest release of ensime-emacs? Make sure to use MELPA to update your packages regularly.
* try nuking your ivy cache (specifically `~/.ivy2/cache/org.ensime`), then type `ensime-update` before trying again (upstream bug in `sbt`).
* check the [tickets flagged as FAQ on the server](https://github.com/ensime/ensime-server/issues?labels=FAQ) and [tickets flagged as FAQ for Emacs](https://github.com/ensime/ensime-emacs/issues?labels=FAQ).
* check the [most recently updated tickets](http://github.com/ensime/ensime-emacs/issues?direction=desc&sort=updated) and do a quick search to see if anybody has reported this already.
* if its about a non-scala specific question, check that the thing you're experiencing isn't from another Emacs mode. If you don't have an Emacs guru to hand, try asking on our gitter channel https://gitter.im/ensime/ensime-emacs

If that solved your problem, great! Please close the ticket. If not, please read on...

There is very little we can do unless you provide us with a **minimal reproduction of the problem**. Creating a standalone project on github is a great way to do this. Please also provide step by step instructions about how you triggered the bug (videos can be useful too). Commit your `.ensime` and the output from the server buffer in the repository. You can also enable `(setq debug-on-error t)` to get a stacktrace in Emacs if an error occurred, which is useful to include.

The next step is for us to help you create a unit test that fails. Once we have an automated failing test, then finding the solution is typically a lot easier.

We'll give as much help as we can, our goal is to grow this community. Honestly, if you're already using Emacs to write Scala code, you have all the skills you'll need to make ENSIME even better. All interactions within the ENSIME community have been friendly and supportive: we want this to continue, formalised by our [Code of Conduct](https://github.com/ensime/ensime-server/wiki/Code-of-Conduct).

fommil:
Ha! I didn't know the auto responder responded to PRs

fommil:
I've hit rebuild. Looks like the failure was transient

fommil:
thanks!

PR #317, opened 2015-12-20
Add basic support for diff refactoring

lxol:
This is a rudimentary support for most of ensime refactoring functionality using a new `diff-refactor` request.  The current prepare-exec-cancel refactoring method is not affected by the change.

I guess we need to add some commands on top of the standard `diff-mode` like applying all hunks, deleting diff file when all hunks applied, etc... What do you think?

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -1072,6 +1072,16 @@ copies. All other objects are used unchanged. List must not contain cycles."
      `(swank:prepare-refactor
        ,proc-id ,refactor-type ,params ,(not non-interactive)) continue)))
 
+(defun ensime-rpc-refactor-diff
+    (proc-id params non-interactive continue blocking)
+  (if blocking
+      (ensime-eval
+       `(swank:diff-refactor
+         ,proc-id ,params ,(not non-interactive)))
+    (ensime-eval-async
+     `(swank:diff-refactor
+       ,proc-id, params ,(not non-interactive)) continue)))
+
 (defun ensime-rpc-refactor-exec (proc-id refactor-type continue)
   (ensime-eval-async `(swank:exec-refactor ,proc-id , refactor-type) continue))
 
@@ -1088,7 +1098,7 @@ copies. All other objects are used unchanged. List must not contain cycles."
 		     continue))
 
 (defun ensime-rpc-async-symbol-designations-for-buffer (start end requested-types continue)
-  (let ((file (cond ((version<= "0.8.18" (ensime-protocol-version))
+  (let ((file (cond ((version<= "0.8.19" (ensime-protocol-version))
 		     (ensime-src-info-for-current-buffer))
 		    (t buffer-file-name))))
     (ensime-eval-async `(swank:symbol-designations ,file ,start ,end ,requested-types)
--- a/ensime-mode.el
+++ b/ensime-mode.el
@@ -75,11 +75,21 @@
       (define-key prefix-map (kbd "C-d q") 'ensime-db-quit)
       (define-key prefix-map (kbd "C-d l") 'ensime-db-list-locals)
 
-      (define-key prefix-map (kbd "C-r r") 'ensime-refactor-rename)
-      (define-key prefix-map (kbd "C-r o") 'ensime-refactor-organize-imports)
-      (define-key prefix-map (kbd "C-r l") 'ensime-refactor-extract-local)
-      (define-key prefix-map (kbd "C-r m") 'ensime-refactor-extract-method)
-      (define-key prefix-map (kbd "C-r i") 'ensime-refactor-inline-local)
+      (define-key prefix-map (kbd "C-r r") (if ensime-refactor-enable-beta
+                                               'ensime-refactor-diff-rename
+                                             'ensime-refactor-rename))
+      (define-key prefix-map (kbd "C-r o") (if ensime-refactor-enable-beta
+                                               'ensime-refactor-diff-organize-imports
+                                             'ensime-refactor-organize-imports))
+      (define-key prefix-map (kbd "C-r l") (if ensime-refactor-enable-beta
+                                               'ensime-refactor-diff-extract-local
+                                             'ensime-refactor-extract-local))
+      (define-key prefix-map (kbd "C-r m") (if ensime-refactor-enable-beta
+                                               'ensime-refactor-diff-extract-method
+                                             'ensime-refactor-extract-method))
+      (define-key prefix-map (kbd "C-r i") (if ensime-refactor-enable-beta
+                                               'ensime-refactor-diff-inline-local
+                                             'ensime-refactor-inline-local))
       (define-key prefix-map (kbd "C-r t") 'ensime-import-type-at-point)
 
       (define-key map ensime-mode-key-prefix prefix-map)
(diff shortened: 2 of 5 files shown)


fommil:
Ok, great! So you're suggesting that we leave it side by side with the current impl until you feel its stable?

Can you please add some tests?

fommil:
we could maybe have some simple rule like less than 10 lines is automatically applied, otherwise the diff pops up in ediff (or something better?) for the user to apply the usual way. (or just always pop up in ediff initially)

fommil:
We could even add magit integration so that refactor won't run unless everything is staged. But maybe that's too heavyweight.

lxol:
Yes I would prefer to have them side by side while we can explore some options of the new functionality.

I'll add tests tater today or tomorrow.

aemoncannon:
Maybe automatic until it's multifile? Same file can be undone with emacs
undo stack

fommil:
point point!

lxol:
I've added 2 tests to the PR. 
The next todo item is automatic diff-apply for single file refactoring?

lxol:
I noticed that ensime-test.el has a mixture of tab and space indentations. Is it a good idea to `whitespace-cleanup` the file?

fommil:
I've been thinking about the UI and reading up on diff-mode, which I believe should be the UI :smile:

Of course, diff-mode is intimidating/confusing for people who don't expect it so I recommend introducing a user setting, nil by default, to use diff-mode. When nil, just apply the diff.

Can you also introduce a temporary beta tester flag that sets this up as the function for all refactoring operations that are supported?

fommil:
An extra flag to automatically apply one file diffs would also be useful!

lxol:
sounds like a plan.

fommil:
Named `ensime-preview-refactor` would be good for the diff-mode thing.

lxol:
changes is the recent update:
- an option to apply hanks automatically,  based on the type of the request, number of files or hunks in the diff file.
- a flag to chose a diff refactory method.

fommil:
it's usually spelled `beta` but I like this, because it reminds me of frank https://www.youtube.com/watch?v=tulFeVPSb7k

fommil:
oh, on the naming, I meant just the `defvar` to be `preview`... it probably makes sense for these new functions to be called `preview` => `diff`. The `defvar` called `ensime-refactor-preview` would be to enable showing the diff in `diff-mode` unless the changes are below the criteria in `ensime-refactor-auto-apply-hunk-limit` (and `ensime-refactor-auto-apply-file-limit`!).

fommil:
but this looks pretty cool! If you can make the renaming changes, this looks good to merge and I'm really keen to try it out. If it seems stable, we'll swap it over to be the default implementation in the next month or so.

lxol:
Regardless of the "auto-apply" criteria we will have the *ENSIME-Refactirng* buffer populated with a diff contents which is effectively a preview of the changes. Do we really need a separate function for that?

lxol:
I will update code tomorrow...

fommil:
@lxol the point of the "preview" feature is to show it to the users before applying it. By default we should just apply it and not show previews (`ensime-refactor-preview nil`), but if the user wants to preview before applying then we meet the criteria before showing it in `diff-mode` for the user to manually apply.

lxol:
@fommil  hopefully this time I get it. A new update has 3 ways of dealing with a refactor request:
 - silently apply changes when `ensime-refactor-preview` is nil

otherwise

- apply and show changes in *ENSIME-Refactoring* buffer when any of the "auto-aplly" criteria matches
- show changes and move focus to *ENSIME-Refactoring* where we can "C-x C-a" to apply hunks manually.
I think the logic is clear from this snippet:

```` lisp
(pcase (list ensime-refactor-preview
             (ensime--refactor-diff-auto-apply-type-p refactor-type)
             (ensime--refactor-diff-auto-apply-file-p diff)
             (ensime--refactor-diff-auto-apply-hunk-p diff))
  (`(nil ,_ ,_ ,_)       (ensime-refactor-diff-apply-silently diff))
  (`(,_ nil nil nil)   (ensime-refactor-diff-preview-popup diff))
  (_                  (ensime-refactor-diff-preview-apply-popup diff)))
````

**Comment by fommil** [comment]
great stuff! I'll try this out over the next week or so of coding :smile:

```elisp
(setq
  ensime-refactor-enable-beta t
  ensime-refactor-preview t
  ensime-refactor-auto-apply-file-limit 1
  ensime-refactor-auto-apply-hunk-limit 1)
```

fommil:
@lxol I just used this for the first time, very cool! :grinning:

I used it for an organise imports and it applied the diff automatically, but it also showed me the diff in a separate buffer. Would it be hard to kill the diff buffer when we auto-apply?

lxol:
@fommil 
> Would it be hard to kill the diff buffer when we auto-apply?

This   is what `ensime-refactor-preview` is for. Set it to nil and diff buffer(preview) won't be created.

fommil:
Oh, that's not what I thought. I thought the preview was to show the diff (so the user can apply it) and then the other settings were to bypass the preview under certain situations.

lxol:
Currently we differentiate preview and auto-apply and we have 3 configuration options:
1. auto-apply (no preview)
2. auto-apply with preview
3. under certain situations do not auto-apply and show preview to apply manually.

What you are saying is that preview should never go with auto-apply and we always auto-apply (option1) unless overriding (option 3). No Option 2.  Was that the original requirements?

fommil:
That sounds good. Sorry for being confusing.

lxol:
@fommil are these options okay? Here  `ensime-refactor-preview` nil and non-nill values can be overridden independently. 
```` elisp
(defcustom ensime-refactor-preview nil
  "If non-nil, Ensime will show a preview of the changes in the
*ENSIME-Refactoring* buffer. Diff hunks can be applied manually
using standard diff-mode commands. Some changes can still be
applied automatically if they satisfy the criteria in
`ensime-refactor-preview-override-hunk',
`ensime-refactor-preview-override-file' or
`ensime-refactor-preview-override-types'.

If nil, Ensime will apply changes automatically unless overridden by 
`ensime-refactor-no-preview-override-hunk',
`ensime-refactor-no-preview-ovverride-file' or 
`ensime-refactor-no-preview-override-types'"
  :type 'boolean
  :group 'ensime-ui)

(defcustom ensime-refactor-preview-override-file 0
  "Override `ensime-refactor-preview' non-nil setting and
  apply changes automatically when the number of affected files
  is fewer than the value"
  :type 'number
  :group 'ensime-ui)

(defcustom ensime-refactor-preview-override-hunk 0
  "Override `ensime-refactor-preview' non-nil setting and
  apply changes automatically when the number of hunks is fewer
  than the value."
  :type 'number
  :group 'ensime-ui)

(defcustom ensime-refactor-preview-override-types  '()
  "Override `ensime-refactor-preview' non-nil setting and
  apply changes automatically when the refactor type is in the
  list.  

Possible types: `organizeImport', `rename', `extract-local',
  `extract-method' or `inlineLocal'."
  :type '(repeat symbol)
  :group 'ensime-ui)

(defcustom ensime-refactor-no-preview-override-file 0
  "Override `ensime-refactor-preview' nil setting and show
  preview when the number of affected files is greater than the
  value"
  :type 'number
  :group 'ensime-ui)

(defcustom ensime-refactor-no-preview-override-hunk 0
  "Override `ensime-refactor-preview' nil setting and show
  preview when the number of hunks is greater than the value."
  :type 'number
  :group 'ensime-ui)

(defcustom ensime-refactor-no-preview-override-types  '()
  "Override `ensime-refactor-preview' nil setting and show
  preview when the refactor type is in the list.

Possible types: `organizeImport', `rename', `extractLocal',
  `extractMethod' or `inlineLocal'."
  :type '(repeat symbol)
  :group 'ensime-ui)


````

fommil:
wow, loads of options! :smile: the semantics sound good though!

lxol:
ok I will implement the logic then...

PR #320, opened 2015-12-30
search for a scala-compiler jar in :scala-compiler-jars 

lxol:
Add support for running scala as an inferior process with an assembly file

### DIFF
--- a/ensime-inf.el
+++ b/ensime-inf.el
@@ -184,8 +184,11 @@ the current project's dependencies. Returns list of form (cmd [arg]*)"
          :java-flags (or (plist-get config :java-flags) ensime-default-java-flags)
          :classpath (ensime--build-classpath
                      (delete-dups (apply #'append
-                                         (ensime--scan-classpath (ensime-read-from-file (ensime--classpath-file (plist-get config :scala-version)))
-                                                                 "\\(scala-compiler\\|scala-reflect\\)\\(-[.[:digit:]]+\\)?\\.jar$")
+                                         (ensime--scan-classpath
+                                          (if (plist-member config :scala-compiler-jars)
+                                              (plist-get config :scala-compiler-jars)
+                                            (ensime-read-from-file (ensime--classpath-file (plist-get config :scala-version))))
+                                          "\\(scala-compiler\\|scala-reflect\\)\\(-[.[:digit:]]+\\)?\\.jar$")
                                          (funcall get-deps config)
                                          (mapcar get-deps (plist-get config :subprojects))))))
       (error "No ensime config available"))))


fommil:
I'm a bit confused about this one, are you using the ENSIME assembly jar to launch the REPL? That's not the intention. We want to keep ENSIME and project jars separate.

There are plans to include the Scala tools jars in the .ensime file, that may be a cleaner solution to what you're trying to solve here (which I assume is because your project doesn't have scala-compiler)

lxol:
My understanding is that `ensime-inf-repl-config` fails regardless of  the REPL launching method. It reads from a non-existent `(ensime--classpath-file scala-version)` file and I was trying to fix that.

fommil:
OK, I see.

I think the proper fix for this should be https://github.com/ensime/ensime-sbt/issues/119

Which means we would parse the project file not the ENSIME server classpath file.

This PR is a workaround but I'm hesitant to merge it because it'll lead to lots of weird classpath errors and ultimately will cost us more time to maintain.

Could you perhaps do the Emacs side fix assuming the correct jars in the .ensime file? Then I'll prioritise updating ensime-sbt to do the right thing.

lxol:
ok, to make sure I understand
- we add a new property to .ensime `:system-deps`
- the ensime-inf fix should search scala-compiler and scala-reflect in `:system-deps` list
- if `:system-deps` property doesn't exist it fallbacks to searching in ensime--classpath-file or assembly-file

fommil:
1. we add a new property to .ensime :scala-compiler-jars
2. the ensime-inf fix should search scala-compiler and scala-reflect in :scala-compiler-jars list
3. if :scala-compiler-jars property doesn't exist it fallbacks to searching in ensime--classpath-file

(i.e. don't use the assembly jar... it'll cause us more pain in the long run if we support it)

lxol:
ok

fommil:
(the reason it'll cause us problems is because it'll add a bunch of ensime jars onto the classpath, and the user will get confused why their libs are the wrong version... e.g. version of shapeless, guava, spray, akka, etc...)

lxol:
should  `:scala-compiler-jars` be per project or subproject?

fommil:
`:scala-compiler-jars` will be a project-level configuration initially. In ensime 2.0, we'll make a lot of this stuff project-specific.

fommil:
@lxol totally off topic, but what is the name of that dired plugin you use for tab expansion?

lxol:
dired-subtree part of dired-hacks https://github.com/Fuco1/dired-hacks
I am not sure if it is in any package repository. I use el-get and install/update packages from github directly.

fommil:
@lxol fyi http://melpa.org/#/dired-hacks-utils

lxol:
@fommil  cool 
I use only few commands from that package. imho filtering is better with projectile/helm.
This is what I found useful:
```` elisp
(define-key dired-mode-map (kbd "i") 'dired-subtree-insert)
(define-key dired-mode-map (kbd "TAB") 'dired-subtree-toggle)
(define-key dired-mode-map (kbd "S-<tab>") 'dired-subtree-cycle)
(define-key dired-mode-map ":n" 'dired-subtree-narrow)
(define-key dired-mode-map ":w" 'widen)
(setq dired-subtree-cycle-depth 100)

````

lxol:
By the way is this PR okay now?

fommil:
oh cool, I didn't even see that you'd updated it!

That looks good, but now I'm confused how we ever got the `scala-library.jar`... it must be because it's always in the project definition.

fommil:
that MELPA package doesn't contain all the files in it. darn.

PR #324, opened 2016-01-01
use the new scala-less assembly jars

### LINKED ISSUE #323: handle new scala-free assembly jars
part of #269 and a response to https://github.com/ensime/ensime-server/pull/1201

fommil:
close #323


fommil:
we don't have an automated test for assembly jar launching, so I tested this manually.

fommil:
that seemed to be a transient failure in the emacs 24.4 tests

PR #349, opened 2016-01-27
Small fixes

### LINKED ISSUE #332: Stop closing my windows - too many ensime functions use (delete-window)
I frequently split my frame in to two windows, sometimes three or four.

Various functions in ensime use (delete-window) in their when their window is no longer to be displayed.

Off the top of my head, *ensime-compilation-result* will kill its window when you hit q, and so does the buffer that appears after invoking (ensime)

In comparison, many built in emacs things (like ibuffer, or buffer-menu, or package-menu) do not affect close windows if they did not create them.

A simple reproduction: open a frame with one window. split that window. invoke M-x ensime. second window disappears.

Desired behaviour - don't close windows that you didn't open. Possible additional behaviour: don't close windows inside process sentinels - this leads to windows magically disappearing when you're doing something else, which is quite disruptive.

### LINKED ISSUE #346: ensime-copy-type-at-point
variant of type-at-point that puts the type into the kill ring, would have been really handy for me today... I'm sure it's trivial to implement.

fommil:
@aemoncannon are you ok if I rename `ensime-print-type-at-point`? and also this closes #346 

@greenrd can you have a look where I remove the global variable? I think that's right, but I'm no elisp expert.

addresses one use of #332

### DIFF
--- a/ensime-inspector.el
+++ b/ensime-inspector.el
@@ -36,12 +36,15 @@
     map)
   "Type and package inspector key bindings.")
 
-
-(defun ensime-print-type-at-point ()
-  "Echo the type at point to the minibuffer."
-  (interactive)
+(defalias 'ensime-print-type-at-point 'ensime-type-at-point)
+(defun ensime-type-at-point (arg)
+  "Echo the type at point to the minibuffer.
+A prefix argument will add the type to the kill ring."
+  (interactive "P")
   (let* ((type (ensime-rpc-get-type-at-point))
          (fullname (ensime-type-full-name-with-args type)))
+    (when arg
+      (kill-new fullname))
     (message fullname)))
 
 (defun ensime-inspector-buffer-p (buffer)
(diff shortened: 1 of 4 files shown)


aemoncannon:
lgtm.

fommil:
@lunaryorn I did a quick google to check if anybody was using this function, it turns out you were, so I added a `defalias` so it doesn't break your config.

lunaryorn:
That's nice, thank you very much, but I'm not actually using the .emacs.d repo anymore; it's just left for archiving.  I'm on spacemacs now and the .spacemacs.d repo is my config now 😎

fommil:
@greenrd I tried it out with the auto generate and it seems to work. I've never actually used `ensime-auto-generate-config`, it's pretty neat. It just occurred to me that it can be used in `.dir-locals.el` which means opt-in per project! :smile_cat: 

FYI this is my `.dir-locals.el` in `ensime-server`

```elisp
((nil . ((yatemplate-owner . "ENSIME Authors")
         (yatemplate-license . "GPL 3.0")
         (ensime-auto-generate-config . t))))
```

and so I don't get pestered about these all the time, I have in my `init.el`

```elisp
(put 'yatemplate-owner 'safe-local-variable #'stringp)
(put 'yatemplate-license 'safe-local-variable #'stringp)
(put 'ensime-auto-generate-config 'safe-local-variable #'booleanp)
```

PR #353, opened 2016-01-31
Migrate to protocol 0.8.20 (no type ids!)

aemoncannon:
This client should be (mostly) backward-compatible with protocol 0.8.19.


aemoncannon:
The light is green, the trap is clean. Merging!

PR #355, opened 2016-02-01
preemptive support for ensime-monkeys

fommil:


fommil:
this should be a no-op, so self merging as its trivial.

PR #363, opened 2016-02-08
Removed setting company-idle-delay as it might be already configured.

iced:


fommil:
This might work for you, but it'll break everybody else's workflow, so I can't accept this PR in its current form.

You could customise the ensime company-mode usage, like I do, if you want to customise further, e.g. https://github.com/fommil/dotfiles/blob/master/.emacs.d/init.el#L640-L644

```elisp
            ;; for small projects, use TAGS for completions
            (make-local-variable 'company-backends)
            (projectile-visit-project-tags-table)
            (setq company-backends
             (if (and tags-file-name
                      (<= 20000000 (buffer-size (get-file-buffer tags-file-name))))
                 '(ensime-company (company-keywords company-dabbrev-code company-yasnippet))
               '(ensime-company (company-keywords company-dabbrev-code company-etags company-yasnippet))))
```

fommil:
http://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Buffer_002dLocal.html

`kill-local-variable` might be all you need, run it after calling `(ensime-mode)`.

fommil:
@aemoncannon was there a reason why we have zero delay in `company-mode` can you remember?

Probably we could do with a less-prescriptive way of using `company`, but this PR is not the way to do it without considering the full impact on the typical user.

iced:
Make configurable var like ensime-company-idle-delay or something? I'm just trying to point you that setting some options for the shared lib, like company, which might be already configured isn't really best idea. And if you do that, you should at least make it easy to undo or disable your customization.

fommil:
@iced we make it a local variable, which won't affect anything other than the `ensime-mode` buffer.

It is trivial to undo this customisation, just use `kill-local-variable` like I suggested above after enabling `ensime-mode`.

If you unset this variable, let us know if you see any unintended consequences. Our integration with `company` involves a lot of hacks because of the nature of the way ensime works.

I agree that we should perhaps investigate and revisit how we set up `company`, but there are good reasons why the hacks are there and we can't just go and remove them without user testing and a solid default alternative that works for most users.

aemoncannon:
@fommil 
Yeah. Our backend is async, so it doesn't really make sense to wait to send the request. IMO we want it to be as responsive as possible, so long as we're not hurting interactivity.

iced:
Tried all "simple" ways:

(add-hook 'scala-mode-hook (kill-local-variable 'company-idle-delay))
(add-hook 'ensime-scala-mode-hook (kill-local-variable 'company-idle-delay))
(add-hook 'ensime-mode-hook (kill-local-variable 'company-idle-delay))

Doesn't work. Of course, what @fommil suggested works, but it's just rewriting part of ensime itself in config :( 

So what about config var like ensime-company-reconfigure-defaults, defaulted to t? Will not change anything for existing users, will allow easy fixes for guys like me.

fommil:
@iced I don't see any problem with customising ensime the way suggested, although I admit it could be cleaner.

I don't think hooks work this way, they are supposed to be lambdas whereas you've passed the value of a function that you've just called.

Try writing your `scala-mode-hook` like my one perhaps?

fommil:
@iced btw, if you'd like to study our company-mode use in more detail to clean this up, we'd be very happy to accept a PR in that area, but it needs to address our unique requirements, rather than pushing that to the user.

e.g.

```
(add-hook 'scala-mode-hook (lambda ()
  (ensime-mode)
  (kill-local-variable 'company-idle-delay)))
```

or (untested)

```
(add-hook 'ensime-mode-hook (lambda () (kill-local-variable 'company-idle-delay)))
```

iced:
Already have first one in my config, 100% the same ;)
And yes, I feel stupid now for (add-hook 'scala-mode-hook (kill-local-variable 'company-idle-delay)) and co. Just realized what it does (way to tired to think ;]). 

OK, I'll try to make nice fix. Please confirm that you don't have SJW CoC, just in case.

fommil:
@iced our CoC is clearly displayed on ensime.org I don't know what the SJW CoC is.

PR #366, opened 2016-02-13
Import qualified name with same base package and group them together

jacobono:
If there is already a statement of `import com.example.MyCoolClass`

and a user is trying to import `com.example.MyOtherCoolClass`,

the result will be `import com.example.{ MyCoolClass, MyOtherCoolClass }`

Prior to this commit, the import statement would have been added
after on a separate line as:

```scala
import com.example.MyCoolClass
import com.example.MyOtherCoolClass
```

## Cask And Ecukes

Additionally, I thought it might be time to use cask and ecukes for (initially) the non async tasks.  I put those changes in a separate commit in case you weren't too keen on it.

In case it isn't obvious, if you like the new non async integration tests, I am volunteering to convert the rest of them to ecukes tests as well in another PR :bowtie: 

And I can also take a look at the async tests eventually after that.

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -1,4 +1,11 @@
-;;; ensime-editor.el  -- Editor and navigation commands
+;;; ensime-editor.el -- Editor and navigation commands -*- lexical-binding: t -*-
+
+;; Copyright (C) 2015 ENSIME authors
+;; License: http://www.gnu.org/licenses/gpl.html
+
+;;; Commentary:
+;;
+;;; Code:
 
 (eval-when-compile
   (require 'cl)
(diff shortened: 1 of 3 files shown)


fommil:
cool! I'll have to have a closer look at this

.drone.yml
```diff
@@ -8,6 +8,7 @@ build:
     - cask pkg-file ;
       cask install ;
       cask build ;
+      cask exec ecukes --no-win;
```
fommil: this fails in our CI

aemoncannon:
Thanks for the import improvements!
Can't say I'm a fan of the ecukes stuff though.

fommil:
yeah, this is really great stuff!

I'm in two minds about the ecukes tests. On one hand they read really well, on the other I can imagine I'd just never remember what I need to type.

.drone.yml
jacobono: I don't know why `--no-win` isn't working for drone, it worked well for me before on travis.  took the `--no-win` off and it works now.

jacobono:
@fommil  @aemoncannon 

At first I wasn't completely sold on ecukes, but I do like 

- the action chains it provides (*that I would think would have to be run on some of the tests*)
- already baked in buffer integration testing 
- how it reads
- tidy output that shows when something fails (IMO the **biggest** improvement over the way tests are currently run)

Sometimes I forget what to type, but I usually go to the `espuds` README to get some of them.

That's just my case for it, but it's your project -- I can remove commit 38dad50 if there is an absolute 

> "I hate this immensely / not a fan / bleh / don't want to try it".

consensus.

I just think it would be cool to use, and easier to use more continuously during development.

There is no reason we couldn't *also* use ERT as well, but I think for integration tests ecukes works better.

fommil:
can we maybe look at using ecukes in a separate PR? My only substantial objection is that our CI build doesn't seem to like it, so I'd like to get to the bottom of that.

jacobono:
> My only substantial objection is that our CI build doesn't seem to like it, so I'd like to get to the bottom of that.

I got it working after all in the `drone.yaml` file.  Apparently how I was used to configuring it in travis is different than what drone.io expects.

```shell
.
AC.
AC.
AC.
AC.
AC.
AC.
AC.
AC.

9 scenarios (0 failed, 9 passed)
55 steps (0 failed, 0 skipped, 55 passed)
```

The other error was from me :open_mouth: in  the `test.sh` run.

The lexical binding introduced in `ensime-editor.el` was causing the undefined local variable in `ensime-util.el` to throw the error.  Easy enough to fix. 

Can still do ecukes in a separate PR if you want though, @fommil.

fommil:
Yeah let's do ecukes separately, possibly widen the scope of things that are tested? This is a great PR and I'm really keen to get it merged.

fommil:
(If you're testing things with ecukes that aren't already tested, we can't exactly say no :-P)

jacobono:
i made the commits so we could separate it if we need to :)

Breaking up the commit into an ecukes branch now and leaving the sole commit.

fommil:
niiice... I used this for the first time and I liked it :smile: @jacobono

PR #367, opened 2016-02-13
Ecukes testing

jacobono:
commit d4a3f61 should go away when #366 is merged into master and I this branch.

@fommil -- do you want widen the scope of testing in this PR to include all the non-async tests?


aemoncannon:
@jacobono Thanks for doing the (extra) work of separating the cukes stuff out!

With deference to the majority, my argument against ecukes is that it puts a translation layer between you and the code. You have to pay for the layer when authoring tests (custom regexes), and when debugging a test failure (what does this test actually do?). This extra work is meant to be worthwhile in exchange for 'readable' tests... but for me it's just not worth it.

greenrd:
I agree. For me it's not worth it either. It's good for non-technical people... but we don't have any of those in this project.

fommil:
I am coming around to them. It solves one problem we have which is that we currently have a monolithic test file and our own test framework (which has caused us a lot of stability problems). These new ecukes tests are more modular than what we have and it seems to be async by default.

I don't think we could reasonably turn down ecukes tests if they are adding to the current suite and increasing the coverage, so perhaps that's where the effort should be spent, instead of translating existing tests.

fommil:
and the second problem it solves is perhaps better failure messages. It's always been a pain to work out what went wrong when something went wrong.

aemoncannon:
Is it possible to use ecukes tests without using english prose?

fommil:
And here's a big one: can the language neutral tests be reused by the other editors? If we could share tests with atom, vim and sublime that would be a huge win

jacobono:
so most of the non async tests don't make sense for ecukes.

Probably a bad branch name - sorry! :scream:   

I just moved all the non async tests to ert.  drone looks good so that's cool :smile: 

The async tests will take a bit longer, but I don't mind looking into them if that is something that would be accepted.

fommil:
wow, this is a big PR! I definitely prefer the ert tests.

The @ensime/atom-panel folks reckon we might be able to reuse cucumber tests, which would be amazing!

If you can convince @aemoncannon and / or @greenrd I think we might have the majority :smile: I think that might mean cucumber tests without so much prose.

fommil:
@jacobono what about alternatives to English prose? Any ideas on that? (don't say Spanish prose...)

hedefalk:
Cool! +1 for cucumber I think.

Awesome if we could separate out cucumber specs to separate repo. Some specs could probably be so general that we can hook into a subset of these from ensime-atom. Atom land uses jasmine but I think we can use https://github.com/cucumber/cucumber-js to hook into Gherkin files… Dunno. Maybe too complex for too little gain, but could be a cool thing.

fommil:
@aemoncannon ! Ping

fommil:
@jacobono ! Ping

fommil:
another data point, people are recommending ecukes for "anything non trivial" https://www.reddit.com/r/emacs/comments/43nh3h/whats_the_best_practice_to_write_emacslispat_2016/

greenrd:
A separate repo for ecukes tests could allow us to write failing tests before we have an implementation (TDD!) and more easily use `git bisect` to find which revision introduced a bug.

OK, I'm coming around to ecukes.

aemoncannon:
+1 If for no other reason than just to be more mainstream :-)
Thanks, @jacobono!

fommil:
@jacobono I think the ayes have it. Can you please squash your commits and ecuke/ert-ise whatever else you feel like :smile:

jacobono:
Wow, I take a weekend break and all the ayes have it!  :sunglasses: 

Thanks for accepting these changes :smile: 

everything is squashed -- all the non async tests are ert/ecukes-ified.

ensime-test.el
```diff
@@ -2560,13 +1921,10 @@ Must run the run-all script first to update the server."
...
   (catch 'done
     (setq ensime--test-had-failures nil)
-    (let ((tests (append ensime-fast-suite
-                         ensime-slow-suite)))
-      (dolist (test tests)
-        (let ((title (plist-get test :title)))
-          (when (integerp (string-match key title))
-            (ensime-run-suite (list test))
-            (throw 'done nil)))))))
+    (let ((title (plist-get ensime-slow-suite :title)))
```
fommil: you dropped a `dolist` here, I'll add it back

PR #368, opened 2016-02-14
Auto-regenerate .ensime for Build.scala

### LINKED ISSUE #350: ensime-auto-generate-config doesn't work for project definitions
if I'm hacking in the `project/` directory, and I don't forget to `gen-ensime-project`, then I can't expect `ensime-auto-generate-config` to do the right thing, which is:

1. if the ensime file is `project/.ensime` run `gen-ensime-project`
2. otherwise run `gen-ensime`

greenrd:
Fixes #350


fommil:
Cool!

PR #371, opened 2016-02-15
return types for ensime-edit-definition

### LINKED ISSUE #358: ensime-edit-definition should return nil/t
with this

```elisp
(defun ensime-edit-definition (&optional where)
  "Lookup the definition of the name at point."
  (interactive)
  (let* ((info (ensime-rpc-symbol-at-point))
         (pos (ensime-symbol-decl-pos info)))
    (if (ensime-pos-valid-local-p pos)
        (progn
          (ensime-push-definition-stack)
          (ensime-goto-source-location pos where)
          t)
      (message "Sorry, no definition found.")
      nil)))
```

I can do

```elisp
(defun ensime-edit-definition-with-fallback ()
  "Variant of `ensime-edit-definition' with ctags if ENSIME is not available."
  (interactive)
  (unless (and (ensime-connection-or-nil)
               (ensime-edit-definition))
    (projectile-find-tag)))
```

which is rather convenient.

fommil:
close #358 

@aemoncannon @greenrd whoever is around. This is a tiny one.

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -271,17 +271,16 @@
   (ensime-edit-definition 'frame))
 
 (defun ensime-edit-definition (&optional where)
-  "Lookup the definition of the name at point."
+  "Lookup the definition of the name at point.
+Goes to the point of the definition (returning point), or fails with `nil'."
   (interactive)
-
   (let* ((info (ensime-rpc-symbol-at-point))
-	 (pos (ensime-symbol-decl-pos info)))
+         (pos (ensime-symbol-decl-pos info)))
     (if (ensime-pos-valid-local-p pos)
-	(progn
-	  (ensime-push-definition-stack)
-	  (ensime-goto-source-location pos where))
-      (message "Sorry, no definition found."))))
-
+        (progn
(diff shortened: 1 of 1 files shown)


fommil:
@greenrd `! Ping`

aemoncannon:
lgtm

ensime-editor.el
```diff
@@ -271,17 +271,16 @@
...
     (if (ensime-pos-valid-local-p pos)
-	(progn
-	  (ensime-push-definition-stack)
-	  (ensime-goto-source-location pos where))
-      (message "Sorry, no definition found."))))
-
+        (progn
+          (ensime-push-definition-stack)
+          (ensime-goto-source-location pos where))
+      (not (message "Sorry, ENSIME couldn't find the definition.")))))
```
aemoncannon: What's this line supposed to do?

fommil: Just legacy compatibility, we always print when we don't find anything. But message returns the message, so we null it.

aemoncannon: Ah hah. Thanks : )

PR #377, opened 2016-02-21
keep trying when there is a process and a port

### LINKED ISSUE #376: support starting up when the port file exists
we recently added support to the server to support starting up against old port files, and that works.

But the Emacs side is unable to handle this and gives the old "check that sbt is on your PATH" error.

fommil:
close #376 

honestly this is a bit hacky, but I keep justifying hacks like this to myself because we'll eventually move to SWANKY over WebSockets.

Yeah, one day.


fommil:
no brainer

Issue #373, opened 2016-02-16
method completion: better default behaviour


fommil:
sometimes I want to insert a method completion but not provide any of the parameters (e.g. if I am changing the method that I'm calling and already have the parameter list)

I think it would be very convenient if the parameters were not inserted if I type the spacebar (like an "empty" hint).

e.g.

```scala
thing.foo(a, b, c)
```

I delete `foo` and complete bar, it looks like this

```scala
thing.bar(a: java.lang.String, b: scala.Int, c: scala.Boolean)(a, b, c)
```

with the point at `a` and the tab jumping enabled for the three parameters. I want, in a single keystroke (space) to go to

```scala
thing.bar(a, b, c)
```

@aemoncannon do you ever get this in Java land too?

aemoncannon:
Yeah for sure, I hit this all the time. I think I filed a bug about it before... Special key could work.

aemoncannon:
Should check what intellij / eclipse does..

fommil:
it's not a real bug report unless it has 6 dupes.

fommil:
I suppose we could auto detect if the next character is `(` or `{` and not expand in that case... in magical unicorn world we'd take a diff of the parameters that are there and what the new requirements and then go to exactly that slot.

aemoncannon:
Yeah. That sounds tough.. but this is suuuuch a bread & butter feature. Challenge accepted? Unfortunately I don't see a good way of leaning on the server for this, as it depends on which completion is chosen, and involves broken syntax. Yay parsing scala in elisp!

aemoncannon:
Data point: Android Studio (based on intellij) doesn't do anything fancy. First off it doesn't insert a param template to begin with.. just shows a help hint in the status bar when you're within the parens. And if you re-type a method name per above, it just duplicates the name with a trailing semi. E.g.

myFun<TAB> => myFunction()
myFun<TAB>myFunction(arg1, arg2) => myFunction();myFunction()

I'm an intellij noob but I don't see any setting for making this fancier.

aemoncannon:
Note: fancier is not necessarily better. Could be they're taking the wise road by getting out of the programmer's way.

fommil:
Winning.

fommil:
(Serious point: I think detecting a `(` in key here)

PR #378, opened 2016-02-23
Delete existing member when completing.

aemoncannon:
Candidate fix for #373. This is a pretty aggressive approach where we replace any ident following (point) with the newly completed ident. Whether or not there is a trailing paren.

@fommil I can't decide if this is right, so I'm going to try it at work some. Will you give it a try as well?

### DIFF
--- a/ensime-company.el
+++ b/ensime-company.el
@@ -178,6 +178,12 @@
       (delete-char (- (length name)))
       (insert to-insert))
 
+    ;; If we're modifying an existing method identifier, delete what
+    ;; was there before and don't mess with the params.
+    (-when-let (suffix (ensime-completion-suffix-at-point))
+	(delete-char (length suffix))
+	(setq skip-params t))
+
     (when (and is-callable call-info (not skip-params))
       (let* ((maybe-braces (ensime-param-section-accepts-block-p
 			    (car (last param-sections))))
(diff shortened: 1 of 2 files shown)


fommil:
Excellent! Will definitely test drive

fommil:
Btw, if we added a dependency on smart parens, we could delete the next block and retain things on the same line.

fommil:
Maybe an optional dep

fommil:
I tried this at work today and didn't notice any difference. Is it doing anything for you? Maybe I loaded it wrong. I'll keep using it all week.

aemoncannon:
oh yeah you should notice right away. i'm guessing it didn't take effect...

fommil:
I still can't see anything. Are you using TAB or RET to make the completion? I use RET.

fommil:
oooh, hold on, I think I do see it. It's only when there is something remaining of the old method name, right? So if I go back and delete the suffix of the existing method, it doesn't do anything. OK, this is pretty subtle and very neat.

fommil:
on the basis that this didn't break anything for me for two whole days of hacking, I think we can say this is ok :smile: it makes completion Do The Right Thing, so it's one of those updates that most users won't notice.

PR #381, opened 2016-02-26
Correct message when using ensime-db-attach

greenrd:
It's necessary to set breakpoints _before_ attaching; neither setting breakpoints nor
running the program is required after attaching.

Also factor some common code out of ensime-db-start and ensime-db-attach.


fommil:
Do you mind please putting dev branches in your repo? It's not a big deal but when I've done this is the past it has dramatically increased the probability of accidentally pushing to master.

greenrd:
OK, will do.

Issue #379, opened 2016-02-24
insert type at point


fommil:
it'd be nice to be able to ask for the type at point and then insert its short name after a colon. It sounds like we might be able to do this without any specific server support, just with regexes / the expand region support.

greenrd:
`C-u C-c C-v t` for this?

fommil:
well that's what I currently do, which involves a lot of manual steps to move to the right place, enter the colon, and paste. In IntelliJ they can do this with a single keybinding.

aemoncannon:
yeah we could have ensime-add-type-annotation. that'd be neat.

aemoncannon:
actually i think i did this in a live-coding ensime talk at some point ;-)

tbje:
I like the idea of add-type-annotation. This one just adds it to the kill-ring:
```
(defun ensime-kill-type-at-point ()
  "Puts the simple name of type at point at the kill ring for easy insertion. A ":" is prepended."
  (interactive)
  (let ((pack-path (ensime-package-path-at-point)))
    (cond ((ensime-visiting-java-file-p)
	   (ensime-inspect-java-type-at-point))
	  (t ;; inspect package if package under point
	   (unless pack-path
	     ;; otherwise, inspect type
	     (let* ((inspect-info (ensime-type-inspect-info-at-point)))
               (kill-new (format ": %s" (plist-get (plist-get inspect-info :type) :name)))))))))
```

fommil:
@tbje is this doing anything more than `C-u C-c C-v t` ?

tbje:
@fommil Ah, didn't know the trick and missed the comment. 

To answer your question, apart from returning the simple / stripped name and the additional ":" no. I guess I could try to improve it a bit so that it actually inserts it in front of the next =. What do you think?

tbje:
```
(defun ensime-add-type-annotation ()
  "Add a type annotation for type at point"
  (interactive)
  (let ((pack-path (ensime-package-path-at-point)))
    (cond ((ensime-visiting-java-file-p)
       (ensime-inspect-java-type-at-point))
      (t 
       (unless pack-path
         (let* ((inspect-info (ensime-type-inspect-info-at-point)))
           (save-excursion
             (search-forward " =")
             (backward-char 2)
             (insert (format ": %s" (ensime-type-name-with-args (plist-get inspect-info :type))))))))))))
```

I only tested it for simple cases. Is there a way to find out if full type is required ?

[...]

PR #390, opened 2016-03-13
Add refactor option to add type annotation to current symbol

d1egoaz:
Add refactor option to add type annotation to current symbol

Mapped to `C-c C-r a`

Fixes https://github.com/ensime/ensime-emacs/issues/379

Demo

Tested with the following examples:
```scala
  private val startTime = startDate.toDateTimeAtStartOfDay.getMillis
  private val endTime = endDate.toDateTimeAtStartOfDay.getMillis

  for {
    a <- List(1, 2)
    b <- Map(1 -> 1)
    (c, d) <- Map("" -> true)
  } yield 1

  def test = 1
  def test2() = 1
  def test3(x: Int) = 1
  def test4(x: Int = 3) = 1
  def test5(x: Int = 3, b: (Int, String) = (1, "")) = 1
  def test6(x: Int = 3, b: (Int, String) = (1, ""))(implicit z: (Int, Boolean)) = 1
  def sum[A](xs: List[A])(implicit m: List[A] ) = m
  def default1(dateStr: String = "a", b: (Int, Int))(implicit x: Int) = { List("") }

  def default(
    dateStr: String = "a", b: (Int, Int)
  )(implicit x: Int) = AuthenticatedAction { request =>
```

### DIFF
--- a/ensime-mode.el
+++ b/ensime-mode.el
@@ -75,6 +75,7 @@
       (define-key prefix-map (kbd "C-d q") 'ensime-db-quit)
       (define-key prefix-map (kbd "C-d l") 'ensime-db-list-locals)
 
+      (define-key prefix-map (kbd "C-r a") 'ensime-refactor-add-type-annotation)
       (define-key prefix-map (kbd "C-r r") 'ensime-refactor-diff-rename)
       (define-key prefix-map (kbd "C-r o") 'ensime-refactor-diff-organize-imports)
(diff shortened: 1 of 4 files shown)


fommil:
@d1egoaz great stuff! Can you please add a test? I recently updated the [contributing](http://ensime.github.io/editors/emacs/contributing/) documentation so it might give some hints.

d1egoaz:
Will do!

Issue #393, opened 2016-03-23
ensime-import-type-at-point strange behaviour without a package declaration


bomgar:
### Steps to reproduce

1. A scala file without a package declaration and no blank lines before the import statements
2. Use ensime-import-type-at-point for a type not imported yet

### Expected behaviour

The import should not be inserted in a new line.

### Actual behaviour

The import statement is inserted in 2 parts. The "i" is the first character in the first line followed by the import statement that was there before. The rest of the statement is in the second line.

fommil:
can you please provide more detail in the production please, e.g. a minimal test project on github.

fommil:
or, better yet, write it up as a failing test in the emacs tests.

bomgar:
In the Hello.scala is an import missing. Try importing it.
https://github.com/bomgar/ensime-emacs-import-test

fommil:
ok, I can see that. If you could minimise a little further and get it into one file, the next step would be to write up a unit test in ensime-emacs. The contributing page gives instructions on how to do that, ask if you need help.

bomgar:
I tried to move everything in the same file. This makes it even worse. The import is inserted in the inner package (which contains the class) instead.
https://github.com/bomgar/ensime-emacs-import-test/tree/one-file

fommil:
Ok, sounds like two separate bugs each requiring a test.

You can reproduce the first bug by importing a class that's not in your project.

Have you read the contributor docs yet? I think this will be easy to fix once you have failing tests in place.

bomgar:
I wrote a test. Now what? Should I create a pull request with the test?

fommil:
Cool! Well now let's try to fix it :-) do you know how to step debug it Emacs lisp the find out which functions are being called for your usecase? Once we know which exact line is causing the problem, the fix might be obvious.

fommil:
ok, looks like `ensime-import-type-at-point` is actually mostly implemented client side (see `ensime-client.el`) so that's where to start.

Note that the whole function only calls out to the ensime server in one place (`ensime-rpc-import-suggestions-at-point`) so you could potentially rewrite your test to be a unit test (without the server) and just mock out this one function. That should speed up the test cycle significantly. We have ert and cucumber style tests, the cucumber ones might be more appropriate for this.

bomgar:
[...]

PR #398, opened 2016-03-28
Fix insert import in files without a package

bomgar:
Inserting an import in a file without a package resulted in a broken
import list. See #393

### DIFF
--- a/features/scala-insert-import.feature
+++ b/features/scala-insert-import.feature
@@ -132,6 +132,32 @@ Feature: Insert Scala Import
     }
     """
 
+  Scenario: Insert in File without package
+    When I open temp scala file "test"
+    And I insert:
+    """
+    import scala.concurrent.duration._
+    class A(value:String){
+      def hello(){
+        new ListBuffer()
+      }
+    }
+    """
+
+    And I go to line "4"
+    And I go to end of line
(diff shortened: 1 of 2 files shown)


features/scala-insert-import.feature
```diff
@@ -132,6 +132,32 @@ Feature: Insert Scala Import
...
     """
 
+  Scenario: Insert in File without package
+    When I open temp scala file "test"
+    And I insert:
+    """
+    import scala.concurrent.duration._
+    class A(value:String){
+      def hello(){
+        println(new /*1*/ListBuffer())
```
fommil: the `/*1*/` is a position indicator for the server tests, it's not needed here.

fommil: probably your line could just be `new ListBuffer` instead of wrapping with `println` and brackets, to better simulate the usecase.

bomgar:
Ok

fommil:
looks like a transient failure in the tests. I've seen that a few times.

fommil:
same transient. Hmm. We might have to track that one down.

fommil:
this is good for merge if you can fix up the test as discussed and squash into one commit (force push)

bomgar:
>  fix up the test as discussed 
did i miss something?

fommil:
nope, you were too quick :smile:

PR #401, opened 2016-03-31
define a key to apply all hunks

lxol:


ensime-refactor.el
```diff
@@ -244,6 +244,14 @@ Do not asks user about each one if `ensime-refactor-save-with-no-questions' is n
         (forward-list))
       (insert (concat ": " shortname)))))
 
+(add-hook 'diff-mode-hook
```
fommil: This will apply to all `diff-mode`s, is that ok?

Maybe adding to `ensime-mode-hook` will be better from a lazy loading perspective.

lxol: makes sense, will do

lxol: I made "a" key biding buffer local. *ENSIME-Refactoring* only.

PR #402, opened 2016-04-01
Skip multiline imports before inserting new import

### LINKED ISSUE #388: import inserted inside multiline curly brace import list
### Steps to reproduce

1. git clone https://github.com/greenrd/ensime-import-bug.git
2. open ensime-import-bug/src/main/scala/ImportBug.scala
3. M-x ensime
4. Wait for ensime to fully start
5. Go to the F in Future
6. C-c C-r C-t
7. Choose scala.concurrent.Future from the completion menu

### Expected behaviour

import inserted as follows:

```
import foo.bar.{Name1, Name2, Name3, Name4, Name5, Name6, Name7, Name8, Name9,
  Name10, Name11}
import scala.concurrent.Future
```

### Actual behaviour

import inserted as follows:

```
import foo.bar.{Name1, Name2, Name3, Name4, Name5, Name6, Name7, Name8, Name9,
  import scala.concurrent.Future
  Name10, Name11}
```

bomgar:
Should fix  #388

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -583,6 +583,19 @@ Returns a function/closure to invoke the necessary buffer operations to perform 
            insert save-excursion)
       (ensime-indent-line))))
 
+(defun ensime-move-to-end-of-multiline-import ()
+  "Move point to the end of a multiline import."
+  (when (not (string-match (rx (zero-or-more any) ?} (zero-or-more whitespace) eol) (thing-at-point 'line)))
+      (goto-char (point-at-eol))
+      (when (not (equal (point) (point-max)))
+        (forward-line 1)
+        (ensime-move-to-end-of-multiline-import)
+      )))
+
+(defun ensime-at-start-of-multiline-block ()
+  "Returns t when at start of multiline block."
+      (string-match (rx (zero-or-more anything) ?{ (zero-or-more (not (any ?})))) (thing-at-point 'line)))
+
 (defun ensime-java-new-import-insertion-decisioning-in-import-block (insertion-range starting-point qualified-name)
   "Search through import statements in buffer above INSERTION-RANGE and STARTING-POINT.
 Decide what line to insert QUALIFIED-NAME."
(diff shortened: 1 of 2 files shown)


ensime-editor.el
```diff
@@ -583,6 +583,19 @@ Returns a function/closure to invoke the necessary buffer operations to perform
...
+  (when (not (string-match (rx (zero-or-more any) ?} (zero-or-more whitespace) eol) (thing-at-point 'line t)))
+      (goto-char (point-at-eol))
+      (when (not (equal (point) (point-max)))
+        (forward-line 1)
+        (ensime-move-to-end-of-multiline-import)
+      )))
+
+(defun ensime-at-start-of-multiline-block ()
+  "Returns t when at start of multiline block."
+      (string-match (rx (zero-or-more anything) ?{ (zero-or-more (not (any ?})))) (thing-at-point 'line t)))
```
fommil: I personally find regex easier to read but I'm ok with this. Have you seen the pcre2el package?

I believe the more standard thing to do is to piggy back off the face highlighting of the buffer but perhaps correctness is more important here than performance.

aemoncannon:
We should really be doing this on the server. 'ensime-rpc-find-import-insertion-point' ?
I'm largely to blame for seeding this import in elisp business, but it's getting silly.

aemoncannon:
(which is not to say that we shouldn't take this much-appreciated fix!)

fommil:
yeah, problem with server side is that there is no source-code preserving AST for scala, so we'd end up just hacking it. Another problem for that GSoC project to solve, and contribute to scala-refactoring! (Assuming we get it)

bomgar:
Handling this on the server side would be much better. Although it would spoil the fun for people working on the implementation for other editors.
Do you want me to change the rx expressions to regex?

fommil:
Nope it's good. I need to look into the build, it's very flaky.

bomgar:
It seems there is a real problem with the old emacs version.

fommil:
If it's something you can fix, that'd be great, otherwise I'll have a look if it's a transient.

fommil:
Oh yeah. I see the error now, yeah looks like a signature changed, you'll have to fix that.

bomgar:
I guess i fixed it. But there is still something wrong with the build.

fommil:
looks like this was a transient

```
Starting test: 'Test debugging scala project.'
ENSIME server starting...
Connecting to Swank on port 38516..
Connected to ENSIME speaking protocol 0.8.20, please wait while the project is loaded.
Initializ
ENSIME ready. case Live(_ : Long) if prosper =>
Location 
attaching to target VM...
Attached to target VM - program running
Resolved 
Installin
Exception on thread 1...
Exception on thread 1...
Reply to canceled synchronous eval request tag=ensime-result-6-sym52430 sexp=(swank:debug-value (:type reference :object-id "372"))
[info] Timeout request received, killing process
```

I'll investigate, but I don't think your stuff is to blame so I'll merge. Thank you!

PR #409, opened 2016-04-04
Compile tests too

lunaryorn:
I've changed `compile` to `test:compile` to make SBT compile the tests as well.  I'm not sure whether that has any unintended side-effects or whether it's even a good idea in general, so I'm opening this PR to "fish for feedback" :sunglasses:

### DIFF
--- a/ensime-sbt.el
+++ b/ensime-sbt.el
@@ -51,7 +51,7 @@
 
 (defun ensime-sbt-do-compile ()
   (interactive)
-  (sbt-command "compile"))
+  (sbt-command "test:compile"))
 
 (defun ensime-sbt-do-run ()
   (interactive)


fommil:
yeah, why not. I'd like to clean up a bunch of these `ensime-sbt` commands at some point. Historically it existed before `sbt-mode` and now pretty much everything is just delegating to `sbt-mode` so this kind of thing should really live there.

fommil:
what I'd like our sbt support to focus on would be the integration between ensime and sbt, and also for extra support for ensime specific commands / tasks (from `ensime-sbt`). e.g. providing convenience functions to allow people to find the current `Project` / `Configuration`. As an example of the kinds of integration that are possible, see https://github.com/ensime/ensime-sbt/issues/181

lunaryorn:
@fommil So you'll merge it?  I'd sign your CA then.

fommil:
yeah, I'll merge it.

lunaryorn:
@fommil Signed.

PR #418, opened 2016-04-15
support typelevel scala releases

### LINKED ISSUE #417: support exotic scala version strings
e.g. `2.11.8-tl-201604131941` currently fails because this fails

```elisp
(version-to-list "2.11.8-tl-201604131941")
```

looks like we need to trim everything past `-` to be able to handle these typelevel scalas.

fommil:
close #417


fommil:
this still doesn't work because the `build.sbt` we create doesn't know how to infer the `scalaBinaryVersion`

Issue #385, opened 2016-03-04
Buffer is read-only: #<buffer *ensime-search-results*>


casertap:
### Steps to reproduce

1. Have a sbt scala project open on emacs and connected ensime 
2. C-c C-v v
3. Type the name of the class, go on the filter window and hit RETURN to go to the source file

### Expected behaviour

I want to see the source of the selected file (The file exists, I have checked)

### Actual behaviour

It shows me the message:
```
Buffer is read-only: #<buffer *ensime-search-results*>
```

fommil:
thanks for following the template, but can you be a bit more specific about the steps? Example project, what is the search term, which one do you select, which file do you expect it to go to?

casertap:
Sorry.
I am inside of this project: https://github.com/RayRoestenburg/akka-in-action/tree/master/chapter-up-and-running

I open the src/main/scala/com/goticks/Main.scala
I set the cursor on line 30 on RestApi class name 
I hit C-c C-v v
I get 2 win windows (1 for filter, 1 to display the result)
I write: ActorSystem
The filter work so I move the cursor to the filter window and select: akka.actor.ActorSystem
I get the message: Buffer is read-only: #<buffer *ensime-search-results*> and nothing happen

fommil:
first two steps are irrelevant, search can happen anywhere. I'll try to reproduce tonight. Did you wait until indexing was completed?

fommil:
what happens if you go to line 27 and type `M-.` to go to the source of the `ActorSystem`. What does `C-c C-c v` print in the minibuffer?

Also, it is unusual to use the mouse to navigate the search results. Try just using the arrow keys (or `p`/`n`) and then using `RET` to select.

casertap:
When I hit ``` M-. ``` on ActorSystem I jump to the end of the Main.scala file.
My minibuffer says: ``` Buffer: *ensime-search-results* ```
I am not using the mouse. To get out of the window where I enter the "ActorSystem" text I use C-x o
Then I use C-p C-n once I am in the search result window but I can not select the file.
I get the same error has before

casertap:
Sorry ``` M-. ``` do works. Thanks
I still have the other problem but that is already good

fommil:
try doing it without `C-x o`

fommil:
if that works, this is potentially a dupe of https://github.com/ensime/ensime-emacs/issues/186

casertap:
[...]

PR #422, opened 2016-04-28
ensime-search: RET key and Highlighting. (#385)

rreckel:
Hi,

This pull request fixes two little bugs in the ensime-search module:

1. The RET key works now when emacs is used in a terminal. "RET" and <return> behave differently when used in a terminal or in a GUI. The fallback for <return> is alway "RET", but "RET" does not fallback to <return> which was not mapped and thus ignored.

2. The highlighting did not work correctly, because the function (ensime-make-overlay) was not used correctly.

Hope this helps a bit. Thanks for this fantastic module!

CU
Roll

### DIFF
--- a/ensime-search.el
+++ b/ensime-search.el
@@ -79,7 +79,7 @@
     (define-key map (kbd "<down>") 'ensime-search-next-match)
     (define-key map "\C-n" 'ensime-search-next-match)
     (define-key map "\C-i" 'ensime-search-insert-import-of-current-result)
-    (define-key map [(return)] 'ensime-search-choose-current-result)
+    (define-key map (kbd "RET") 'ensime-search-choose-current-result)
     map)
   "Keymap used by ensime-search.")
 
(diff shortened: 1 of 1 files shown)


fommil:
cool, thanks!

ensime-search.el
```diff
@@ -285,7 +285,7 @@
 	(goto-char target-point)
 	(setq ensime-search-selection-overlay
 	      (ensime-make-overlay target-point (point-at-eol)
-				   nil 'ensime-warnline-highlight))
+				   nil '(:face 'ensime-warnline-highlight)))
```
fommil: might not have needed the `'` on `'ensime-warnline-highlight` but if it works then no big deal.

rreckel: Oups... Didn't know that! Sorry!
I'm new to elisp, but I will try to improve fast ;-)
After all, I am really impressed by the capabilities of Emacs and Ensime!!
Elisp is on the best way in getting my second favorite language after Scala....!

PR #424, opened 2016-05-03
Added gutter icons for compile errors and warnings in terminal mode

rreckel:
Hi,

I like the icons that show the errors and warnings in the fringe. As I work a lot in terminal mode, it was annoying that these icons did not work, because there is no fringe in terminal mode.

So I made some changes to show the icons in the left margin.

By default, the icons are now shown, but they can be switched of, because the left-margin is also used by git-gutter or linum and they could conflict with these modes. 

What do you think?
Is it worth integrating?

Anyway, I think it should not interfere with the 'normal' window-system mode ;-)

Roll

### DIFF
--- a/ensime-mode.el
+++ b/ensime-mode.el
@@ -255,6 +255,9 @@
         (add-hook 'after-change-functions
                   'ensime-after-change-function nil t)
 
+        (add-hook 'window-configuration-change-hook
+                  'ensime-show-left-margin-hook)
+
         (ensime-idle-typecheck-set-timer)
 
         (when ensime-tooltip-hints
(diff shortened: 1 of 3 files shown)


rreckel:
Oups.... The test run was killed due to a timeout....
Did my changes cause this error?

Could you please tell me how to run the tests locally?
Thanks

fommil:
details should be in the contributing page on ensime.org

rreckel:
Ok. Got it running now.
The cask test run ok with no errors, but the run_emacs_tests.sh
stop at 'Test debugging scala project.' with
error in process filter: Assertion failed: (= 0 (apply (quote call-process) ensime-sbt-command nil "*sbt-test-compilation*" nil (quote ("compile"))))

Is there something wrong with my installation??

Anyway the tests passed now on the server ;-)

rreckel:
Closed the PR by accident, Sorry

fommil:
sorry I totally forgot to review this, remind me at the weekend if it still goes unviewed.

rreckel:
No stress. I am in vacation 'til Monday anyway ;-)

fommil:
thanks!

PR #425, opened 2016-05-09
Changed TAB keybinding to work in terminal mode

rreckel:
I changed the local key binding of the "TAB" key, because it didn't work in terminal mode.

### DIFF
--- a/ensime-company.el
+++ b/ensime-company.el
@@ -125,7 +125,7 @@
   (yas-minor-mode-on)
   (set (make-local-variable 'company-idle-delay) 0)
   (set (make-local-variable 'company-minimum-prefix-length) 2)
-  (local-set-key [tab] 'ensime-company-complete-or-indent))
+  (local-set-key (kbd "TAB") 'ensime-company-complete-or-indent))
 
 (defun ensime--yasnippet-complete-action (&optional candidate-in force-block)
   "If the candidate is a callable symbol, expand a yasnippet template for the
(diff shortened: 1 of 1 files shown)


ensime-company.el
```diff
@@ -125,7 +125,7 @@
   (yas-minor-mode-on)
   (set (make-local-variable 'company-idle-delay) 0)
   (set (make-local-variable 'company-minimum-prefix-length) 2)
-  (local-set-key [tab] 'ensime-company-complete-or-indent))
+  (local-set-key (kbd "TAB") 'ensime-company-complete-or-indent))
```
fommil: I use a combination of `[tab]` and `"<tab>"` in my `.emacs`. Can you please point to some docs that describe the different? I don't want to break it for everybody in the various Windowed OSes.

rreckel: The problem is that in tty the physical TAB key is bound to C-i and "TAB".
To distinguish C-i and "TAB" in the window-system, another key was introduced <tab> which doesn't translate to C-i. (freeing this combination)
A better explanation can be found in: https://www.emacswiki.org/emacs/TabKey
and of cause the docstrings of edmacro-mode:

```
 * The special words RET, SPC, TAB, DEL, LFD, ESC, and NUL represent
   special control characters.  The words must be written in uppercase.

 * A word in angle brackets, e.g., <return>, <down>, or <f1>, represents
   a function key.  (Note that in the standard configuration, the
   function key <return> and the control key RET are synonymous.)
   You can use angle brackets on the words RET, SPC, etc., but they
   are not required there.
``` 
So, you are right, if somebody did bind a command to C-i, this fix could potentially break the system. (Perhaps Evil mode does such bindings,.....?)

Could this be corrected with a (if (window-system) ...)
What do you think?

fommil:
yeah, that sounds like a good workaround

PR #426, opened 2016-05-09
[discussion needed] Fix hanging import type at point when type does not exist

mriehl:
When I have a typo and try to import a type at point, emacs freezes and I have to `<C-g>` out.

Local debugging showed that the `suggestions` are a list with `nil` in it (`(nil)`).
So the `when` check doesn't work.

I am an elisp noob, so I am not sure that this is the correct fix. In scala I would not create a `List` or `Seq` with `null` in it, so perhaps this is the underlying problem that actually needs to be addressed?

I'm also unsure how to test this. I'd like to have a feature entry that sets the cursor on a word and then calls the related function. But I'd have to assert that the autocomplete list is not shown (nothing should happens and emacs should not hang), and I don't know how to do that. Happy to hear out any suggestions or pointers.

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -694,7 +694,7 @@ Decide what line to insert QUALIFIED-NAME."
 	 (name-start (plist-get sym :start))
 	 (name-end (plist-get sym :end))
 	 (suggestions (when name (ensime-rpc-import-suggestions-at-point (list name) 10))))
-    (when suggestions
+    (when (car-safe suggestions)
       (let* ((names (mapcar
 		     (lambda (s)
 		       (propertize (plist-get s :name)


fommil:
cool! This sounds sensible, but typically we'd want a test to go around this (anything that isn't tested is prone to removal in a later refactor).

Before you do that, have a look at the server events response. There is info in the Contributing page about how to capture the events. Is the problem there, or in later processing?

fommil:
I'm merging this early because it is an obviously good workaround, but it would be good to continue the investigation and get to the bottom of it.

mriehl:
I'll definitely look into it as soon as I am able using the information from the contributing page. Thanks for merging and for your suggestions!

mriehl:
@fommil : I think I've got it figured out, learned a thing or two about ensime and elisp in the process :grinning: 

The ensime-server needs to `flatMap` when producing the import suggestions. This currently generates an `ImportSuggestions(List(List(...))` which leads to the convoluted sexp handling in `ensime-emacs`.

Using a snapshot `ensime-server` and custom `ensime-emacs` works without the `car-safe` hack for me. I can also confirm this with an ecukes test.

But now my problem is that this change is not backwards-compatible. It also probably breaks other editors. So I'm not sure how to go about it.
One way would be to duplicate the function and slowly migrate away from the current implementation. 
I could also try to get all editor plugins to handle both cases first, then change the server. But this would still cause issues to people who upgrade the server before upgrading their editor plugin.
Is there some kind of infrastructure that allows me to see which editors I would break with an ensime-server change?
Or maybe this change is too much to do at once and the current hack is preferred?

Thank you for your time!

fommil:
I think if the client is receiving what is defined by the api then all is an intended. It is difficult to justify changing the api but there is the possibility to do so in ensime 2.0 when we tidy up the domain objects.

PR #432, opened 2016-05-16
Use helm for import-type-at-point

bomgar:
Hi.

i want to integrate helm for some ensime features. To test if it works i use helm for import-type-at-point. I'm not sure if I should continue working on it. Is this something worth pursuing?

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -13,6 +13,9 @@
 
 (require 'dash)
 (require 'popup)
+(require 'ensime-vars)
+
+(autoload 'ensime-helm-select-entry "ensime-helm")
 
 (defvar ensime-compile-result-buffer-name "*ENSIME-Compilation-Result*")
 
@@ -685,6 +688,13 @@ Decide what line to insert QUALIFIED-NAME."
       (goto-char (point-at-bol))
       (funcall (funcall insert-import-fn insertion-range starting-point qualified-name)))))
 
+(defun ensime-ask-user-to-select-entry (title entries)
+  "Prompts the user to select an entry of entries"
+  (if ensime-use-helm
+      (ensime-helm-select-entry entries title)
+      (popup-menu* entries :point (point))
+    ))
+
 (defun ensime-import-type-at-point (&optional non-interactive)
   "Suggest possible imports of the qualified name at point.
  If user selects and import, add it to the import list."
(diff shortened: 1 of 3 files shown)


fommil:
the build will probably fail because we don't depend on helm. It is important that helm is added as an optional, not required, dependency. You can do this in the `Cask` file.

ensime-helm.el
```diff
@@ -0,0 +1,36 @@
+;;; ensime-helm.el
+;;
+;;;; License
+;;
+;;     Copyright (C) 2015 Aemon Cannon
```
fommil: pretty sure Aemon didn't write this :smile:

Please follow the header pattern from `ensime.el` and

fommil:
generally, I like the idea of adding support for helm. Whether or not it is best in `ensime` or a separate `ensime-helm` (or `helm-ensime`) package is up for debate.

Certainly we'd need some tests here. Our approach is that if a change doesn't break a test, it's good.

fommil:
looks like you're out of date with master.

fommil:
oh cool, you did make it optional. Just no tests.

bomgar:
> Whether or not it is best in ensime or a separate ensime-helm (or helm-ensime) package is up for debate.

I like the idea of a separate package. That would require some changes in the ensime code-base. In this case opening the popup menu should be done in a function so it can be replaced by the one that uses helm.

One the other hand for small things like this a extra package is probably to much.

bomgar:
I'm nor sure how to write a test for this. Any pointers?

fommil:
have you seen the contributing docs on http://ensime.org yet?

bomgar:
It seems to me all existing tests "cheat" and use the non-interactive import method. I was hoping to find a test for the current popup behavior. But either it doesn't exist or I can't find it.

ensime-editor.el
```diff
@@ -685,6 +688,13 @@ Decide what line to insert QUALIFIED-NAME."
       (goto-char (point-at-bol))
       (funcall (funcall insert-import-fn insertion-range starting-point qualified-name)))))
 
+(defun ensime-ask-user-to-select-entry (title entries)
+  "Prompts the user to select an entry of entries"
+  (if ensime-use-helm
+      (ensime-helm-select-entry entries title)
+      (popup-menu* entries :point (point))
+    ))
```
fommil: I know we're inconsistent, but we're trying to use standard lisp conventions of no dangling brackets. Also, I believe the standard formatted form for `if` is

```elisp
(if test
    then
  else)
```

fommil:
I'll merge this as I'm keen to encourage this angle of thought. Can you please send an update to the website docs? And also mention on the gitter channel / tweet (and let me know to retweet).

bomgar:
Pull request for the documentation https://github.com/ensime/ensime.github.io/pull/125

Sorry but I don't have a twitter account.

PR #433, opened 2016-05-16
Allow the popup to be replaced by something else

bomgar:
As discussed in https://github.com/ensime/ensime-emacs/pull/432 it might be a good idea to abstract the user interaction for selecting things. This allows the user to replace the import-type-popup with helm or something similar.

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -65,6 +65,12 @@
   "Key bindings for the uses popup.")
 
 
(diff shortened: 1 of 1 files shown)


bomgar:
I'm not sure if the variable is the right choice here. One could also provide a `import-type-at-point` function that takes the popup method as parameter.

fommil:
I don't know, I think I'd rather see the helm support added into `ensime` until it gets big enough to split out. I'm never keen on complicating the codebase.

PR #442, opened 2016-05-18
quick fix for broken unit tests

fommil:
I found a strand of code in here that looks like it needs to be unravelled and removed, but this is the quick fix.

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -858,13 +858,13 @@ copies. All other objects are used unchanged. List must not contain cycles."
   (ensime-eval
    `(swank:doc-uri-at-point ,file ,(ensime-externalize-offset point))))
 
-(defun ensime-rpc-doc-uri-for-symbol (fqn &optional member-name member-signature)
+(defun ensime-rpc-doc-uri-for-symbol (scala-name &optional member-name member-signature)
   (ensime-eval
-   `(swank:doc-uri-for-symbol ,fqn ,member-name ,member-signature)))
+   `(swank:doc-uri-for-symbol ,scala-name ,member-name ,member-signature)))
 
-(defun ensime-rpc-symbol-by-name (fqn &optional member-name member-signature)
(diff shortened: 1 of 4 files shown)


aemoncannon:
What was the bug out of curiousity?

fommil:
oh I introduced it with the latest server changes. Technically the protocol didn't change, but the interpretation of some Strings is now more well defined. e.g. `name` is now the scala type name with the type args, whereas before there were assumptions that the type params needed to be added.

This was in the other side, where we created a fake `type-info` instance out of a user provided string on the client side, but didn't follow the same standards.

Probably we can just remove all the functions, they tend to get converted into Strings at some point anyway and are not being used to communicate with the server.

fommil:
@aemoncannon I'd be interested to know if these changes improve the method finding in Java code! Now that we're using real FQNs with method descriptors, it should really improve FQN => Symbol lookup in the analyzer as well as FQN => indexer lookup. I didn't touch the Java code, so I may have broken something in there.

fommil:
@aemoncannon btw, off topic, is the Java 7 filewatcher stuff working for you on your work project? I'm seeing incredible results on my (gigantic) work project. I used to just turn it off, now it basically uses 0% CPU and reacts instantly.

aemoncannon:
@fommil btw historical context:  This is fall out from the removal of type-ids. Specifically, for completions the parsing allowed me to handle a couple cases (summary signatures when scrolling, and completion expansion) without bloating the completion response and without adding followup RPCs.

aemoncannon:
@fommil can't wait to try out the new changes! (been somewhat away from java code last few weeks). Will try soon : )

fommil:
@aemoncannon heh, well I suspect I probably broke all the things with that latest code dump, but it was needed. We're hoping to cut 1.0 pretty soon and then move onto Java 7.

Thanks for the background. I'll have to study exactly what that client side parsing is doing before I make any changes.

fommil:
oh god, I hope that doesn't mean you're ending up in **management**!!

PR #451, opened 2016-05-28
Use helm to list search results

bomgar:
Provides an alternative to ensime-search (optional)

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -995,6 +995,14 @@ copies. All other objects are used unchanged. List must not contain cycles."
      ,max-results
      )))
 
+(defun ensime-rpc-public-symbol-search
+  (names max-results)
+  (ensime-eval
+   `(swank:public-symbol-search
+     ,names
+     ,max-results
+     )))
+
 (defun ensime-rpc-async-public-symbol-search
   (names max-results continue)
   (ensime-eval-async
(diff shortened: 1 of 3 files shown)


fommil:
This is all great stuff but you're going to have to start thinking about testing. Sooner or later were going to remove everything that doesn't have a test. Can you have a think about that please?

bomgar:
I'm happy to provide some tests but I need help to do that. It is hard to use the current tests as a template.

fommil:
I'd recommend using the cucumber test framework, look under `feature` for some examples, failing that use `ert` (see `test/ensime-emacs-tests.el`). We want to get away from using server integration tests for everything so I recommend learning how to use a mocking framework like https://github.com/rejeep/el-mock.el  or (complementary) https://github.com/nicferrier/emacs-fakir

More generally, it would be really fantastic if we were able to rewrite all our tests to the cucumber with mocking style, then only have one or two smoke tests that launch and connect to the real server.

fommil:
I'm happy for you to do that in a follow up PR, keep it coming! :smile:

ensime-client.el
```diff
@@ -995,6 +995,14 @@ copies. All other objects are used unchanged. List must not contain cycles."
      ,max-results
      )))
 
+(defun ensime-rpc-public-symbol-search
```
fommil: we probably need to do this for all the API calls, abstract the API from the implementation because it **will** change.

fommil:
any chance of a GIF of this working? I'd love to see it! I'd really like to move search over to ido or something lighter weight by default.

bomgar:
I'll have a look at el-mock.

bomgar:
The colors are a little bit off though
ensime_helm_search

PR #453, opened 2016-05-30
remove client side name parsing

### LINKED ISSUE #449: remove client side scala name parsing
remove `ensime-call-completion-info` and replace it with something that just reads the values from the wire as per https://github.com/ensime/ensime-server/issues/1467

`ensime-brief-type-sig` also needs to go then we can rip out the entire backing `ensime-parse-type-info-from-scala-name` thorn.

This may end our support for auto-complete-mode so anybody who uses it might need to copy that file locally.

fommil:
close #449

### DIFF
--- a/ensime-macros.el
+++ b/ensime-macros.el
@@ -23,6 +23,14 @@
 
 (require 'cl-lib)
 
+(defmacro ensime-plist-bind (args expr &rest body)
+  ;; http://emacs.stackexchange.com/questions/22542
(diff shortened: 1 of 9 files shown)


ensime-macros.el
```diff
@@ -23,6 +23,14 @@
 
 (require 'cl-lib)
 
+(defmacro ensime-plist-bind (args expr &rest body)
```
fommil: @aemoncannon you might like this one

Issue #458, opened 2016-06-01
Symbol's value as variable is void: to-insert


fommil:
I am seeing some failures in `ensime--annotate-completions` with completion payloads that looks a little bit like #454 (which is missing a reproduction).

I was able to get the problem by typing

``` elisp
(ensime-plist-bind
      (type-info is-callable to-insert name)
      '(:type-info (:arrow-type t :name "(String) => PathFinder" :full-name "(java.lang.String) => sbt.PathFinder" :result-type (:arrow-type nil :name "PathFinder" :decl-as class :full-name "sbt.PathFinder") :param-sections ((:params (("literal" (:arrow-type nil :name "String" :decl-as class :full-name "java.lang.String")))))) :name "/" :type-sig (((("literal" "java.lang.String"))) "sbt.PathFinder" nil) :is-callable t :relevance 80)
      (propertize name
                  'symbol-name name
                  'type-info type-info
                  'is-callable is-callable
                  'to-insert to-insert))
```

but then it went away! I'm guessing it's got something to do with emacs symbol interning or file opening order.

fommil:
or maybe the `cl-destructuring-bind` is buggy and we should use use `-let` instead.

fommil:
the equivalent call from 454 would be with this payload:

```elisp
 ((:name "match" :type-sig (nil "" nil) :relevance 100) (:name "map" :type-sig (((("f" "scala.Function1[scala.Any, scala.concurrent.S]")) (("executor" "scala.concurrent.ExecutionContext"))) "scala.concurrent.Future[scala.concurrent.S]" t) :is-callable t :relevance 90) (:name "mapTo" :type-sig (((("tag" "scala.reflect.ClassTag[scala.concurrent.S]"))) "scala.concurrent.Future[scala.concurrent.S]" t) :is-callable t :relevance 90)))
```

timcharper:
I upgraded from `20160521.1546` to `20160602.1103` via ELPA and this broke similarly. Then, I switched back and it worked again.

Then, I switched to a cloned version of `ensime-emacs`, caf2e8b7a57a604c1a2a4d999a1e963837a75bd3, and that also works, even though this appears to be the same code as contained by `20160602.1103`.

```
Debugger entered--Lisp error: (void-variable to-insert)
[...]

Issue #463, opened 2016-06-11
company-mode templates should use short type names


fommil:
the long names are distracting, use short type names.

Issue #321, opened 2015-12-31
:paste may be better than :load in ensime-inf-load-file


prateemmandal:
I got into trouble while loading this code into scala> prompt using C-c C-v l (code and error is in the link below)
http://lpaste.net/148133

In the IRC chat room someone helpfully suggested to try out :paste which worked for me. Then I looked at what C-c C-v l was bound to and found that :load is used instead of :paste in ensime-inf-load-file. Then while searching the difference between :load and :paste I found this link
http://blog.japila.pl/2013/04/when-in-scala-repl-paste-might-be-your-better-friend-than-load/

There may be reasons because of which :load may be preferred. If not will it be better to change it to :paste?

fommil:
Thank you for using ENSIME!

Most tickets can be resolved easily by following a simple process.

Before one of the core devs can respond to this ticket, please confirm the following:

1. you nuked ENSIME from your ivy cache (i.e. `rm -rf ~/.ivy2/cache/org.ensime`).
2. you are using the latest release of `ensime-emacs` (i.e. update `ensime` via MELPA).
3. you are using the latest release of `ensime-server` (i.e. type `M-x ensime-update`).
4. you checked the [tickets flagged as FAQ for Emacs](https://github.com/ensime/ensime-emacs/issues?labels=FAQ).
5. you checked the [tickets flagged as FAQ on the server](https://github.com/ensime/ensime-server/issues?labels=FAQ).
6. you checked the [most recently updated tickets](http://github.com/ensime/ensime-emacs/issues?direction=desc&sort=updated)
7. you checked `M-x customize-group ensime` (or viewed `ensime-vars.el`) to see if this is already handled.
8. (if you are using spacemacs) you asked on the spacemacs user list first.

If that solved your problem, great! Please close this ticket. If not, please read on...

There is very little we can do unless you provide us with a **minimal reproduction of the problem**. Please provide:

1. a standalone project on github (or use `ensime-server`)
2. step by step instructions to trigger the bug (videos are useful)
3. attach your `.ensime` to the repository (or in a gist)
4. also attach the output from the ENSIME server buffer
5. also attach an emacs stacktrace (if relevant), e.g. via `(setq debug-on-error t)`

**The next step is for us to help you create a unit test that fails.** We'll give as much help as we can, our goal is to grow this community. You have all the skills you'll need to make ENSIME even better.

[...]

PR #468, opened 2016-06-15
Fixes bug where to-insert was void.

jmpage:
Fixes bug where to-insert was void in `ensime--annotate-completions`. This was done by switching back to using `let` from `ensime-plist-bind`, as suggested by @fommil. I'm pretty new to lisp, so please let me know if there are any changes that you would like; I would be happy to make them!

Should resolve ensime/ensime-emacs#458 (at least it seems to have fixed the issue for me on my previously buggy install of ensime).

### DIFF
--- a/ensime-completion-util.el
+++ b/ensime-completion-util.el
@@ -64,14 +64,13 @@
  being passed through the innards of auto-complete or company."
   (mapcar
    (lambda (completion)
-     (ensime-plist-bind
-      (type-info to-insert name)
-      completion
-      (propertize name
-                  'type-info type-info
-                  'to-insert to-insert)))
-   completions))
-
+     (let* ((name (plist-get completion :name))
+            (type-info (plist-get completion :type-info))
+            (to-insert (plist-get completion :to-insert)))
+       (propertize name
+                   'type-info type-info
+                   'to-insert to-insert
+                   ))) completions))
 
 (defun ensime-completion-prefix-at-point ()
   "Returns the prefix to complete."


fommil:
Hmm, I'd rather not do it this way. This was a PoC to understand how we could use destructuring throughout the codebase. I want to understand why this fails before we do anything with it. I'd perhaps be open to moving to -let to see if it has the same problems, but not all the way back to let... that's admitting defeat.

Nice try :-)

fommil:
@jmpage but since you're interested in lisp, I've got a tonne of other tickets that would be perfect for a lisp beginner! I've just had a look at the Low Hanging Fruit tag, and some of those labels need to be removed (this one definitely isn't an easy one for example)... how about #463 ? It should just be a case of changing `:full-name` to `:name` in `ensime-company` but I haven't had the time to send a PR.

fommil:
(or change this code to use `destructuring-bind` directly to see if that fixes it... then we can be reasonably confident that the bug is in my macro)

fommil:
or this one, which would really help out people who use the REPL a lot (I don't tend to use it so this sort of thing goes to the bottom of my priority list) https://github.com/ensime/ensime-emacs/issues/321

fommil:
(if you want to go for using `destructuring-bind`, just push to this branch and let me know to reopen the PR)

jmpage:
Thanks for the feedback, I'll poke around more tonight and see what I can do with `destructuring-bind`.

fommil:
I only found out about it recently when reading http://www.gigamonkeys.com/book/ so it is very possible that I made a mistake somewhere.

Ideally what I'd like is if we used `defclass`es for the full API and then used slot destructuring (which I also just learnt about last week), but that's quite a lot of work. Possibly worth it when we move over to the new websockets API for ensime 2.0.

PR #469, opened 2016-06-15
Use short type names in company-mode templates.

jmpage:
Resolves ensime/ensime-emacs#463


fommil:
Thanks! Everybody will notice this one.

fommil:
(also Vim fixed this last week, and we can't possibly let them have anything we don't have)

jmpage:
I completely agree :wink:

fommil:
this has been released http://melpa.org/#/ensime

PR #471, opened 2016-06-17
Remove special casing for by-name parameters

iravid:
This functionality has been made redundant by fixing ensime/ensime-server#1477.

### DIFF
--- a/ensime-company.el
+++ b/ensime-company.el
@@ -54,23 +54,22 @@ block notation for the final parameter."
              (let* ((type-info (cadar params))
                     (block-params (plist-get (car (plist-get type-info :param-sections)) :params))
                     (result-type (plist-get type-info :result-type)))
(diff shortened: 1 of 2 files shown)


iravid:
Hey @fommil, I think this should do the trick, but not sure how to test locally against ensime-server 2.0.

fommil:
sorry, this won't work. You need to support the old names as well as the new names. We currently don't have a client that is 1.9+ protocol only.

fommil:
to run tests locally, see the `.drone.yml` of the ensime-server project. `ENSIME_SERVER_VERSION` is used, but my main concern is compatibility with 1.0.0

iravid:
Agreed that this won't work - closing this out to find a better approach. Thanks!

fommil:
@iravid how about just changing the regex to handle (I'm using pcre here, so translate into emacs regex) `^(=>|<byname>).*` ?

Issue #423, opened 2016-05-01
support for ensime-sbt's ensimeCompileOnly task


fommil:
https://github.com/ensime/ensime-sbt/issues/181

fommil:
@jmpage this is also another easy-ish one. The idea here would be to create a command that:

1. discovers the name of the subproject that the current buffer is in
2. creates a command such as `subproject/compileOnly /full/path/to/current/Buffer.scala` (it should be possible to add a test for this, you might need to mock out some ensime function that returns the config for the file)
3. sends to `sbt-command`
4. add a keyboard shortcut and menu entry under the `C-c C-b` (sbt) namespace.

We'll probably rename `compileOnly` to `ensimeCompileOnly` but I'll take care of renaming the command when I do that.

jmpage:
Thanks for the suggestion. I think I got a handle on what I need to do. Hopefully I'll have a PR ready soon!

fommil:
FYI I renamed this to `ensimeCompileOnly` in sbt-ensime 0.6.

PR #476, opened 2016-06-19
ensimeCompileOnly support

jmpage:
Resolves: ensime/ensime-emacs#423

Please let me know if changes are needed.

### DIFF
--- a/ensime-sbt.el
+++ b/ensime-sbt.el
@@ -53,6 +53,13 @@
   (interactive)
   (sbt-command "test:compile"))
 
+(defun ensime-sbt-do-compile-only ()
+  (interactive)
+  (let ((subproject (ensime-sbt-find-subproject buffer-file-name)))
+    (if subproject
+        (sbt-command (concat subproject "/ensimeCompileOnly " buffer-file-name))
+      (sbt-command (concat "ensimeCompileOnly " buffer-file-name)))))
+
 (defun ensime-sbt-do-run ()
   (interactive)
   (sbt-command "run"))
--- a/ensime-sbt.el
+++ b/ensime-sbt.el
@@ -144,14 +144,13 @@ again."
                                       (?q "test-[q]uick" "testQuick"))))
     (concat module "/" source-set task)))
 
-(defun ensime-sbt-find-subproject (file-name source-set)
-  (when source-set
-    (let* ((config (-> (ensime-connection) ensime-config))
-           (subprojects (plist-get config :subprojects))
-           (matches-subproject-dir? (lambda (dir) (string-match-p dir file-name)))
-           (find-subproject (lambda (sp)
-                              (-any matches-subproject-dir? (plist-get sp :source-roots)))))
-      (-> (-find find-subproject subprojects) (plist-get :name)))))
+(defun ensime-sbt-find-subproject (file-name)
+  (let* ((config (-> (ensime-connection) ensime-config))
+         (subprojects (plist-get config :subprojects))
+         (matches-subproject-dir? (lambda (dir) (string-match-p dir file-name)))
+         (find-subproject (lambda (sp)
+                            (-any matches-subproject-dir? (plist-get sp :source-roots)))))
+    (-> (-find find-subproject subprojects) (plist-get :name))))
 
 (defun ensime-sbt-test-dwim (command)
   (let* ((file-name (or buffer-file-name default-directory))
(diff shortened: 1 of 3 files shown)


ensime-sbt.el
```diff
@@ -53,6 +53,14 @@
   (interactive)
   (sbt-command "test:compile"))
 
+(defun ensime-sbt-do-compile-only ()
+  (interactive)
+  (let* ((file-name (or buffer-file-name default-directory))
```
fommil: I'm not sure the SBT command can handle directories

jmpage: right!

ensime-sbt.el
```diff
@@ -53,6 +53,14 @@
...
+         (subproject (ensime-sbt-find-subproject file-name)))
+    (if subproject
```
fommil: There might be a fancy way of doing this in the `s` library.

jmpage: Will look into this.

ensime-sbt.el
```diff
@@ -53,6 +53,13 @@
   (interactive)
   (sbt-command "test:compile"))
 
+(defun ensime-sbt-do-compile-only ()
+  (interactive)
+  (let ((subproject (ensime-sbt-find-subproject buffer-file-name)))
+    (if subproject
+        (sbt-command (concat subproject "/ensimeCompileOnly " buffer-file-name))
```
fommil: It might make it easier to test this code if the calculation of the command was separated from the running of it with sbt-command (you'd need to provide a .ensime file and probably mock the connection). That would mean you wouldn't need to write a big slow integration test.

fommil:
LGTM This is looking pretty good! If this goes green, feel free to merge when you're happy with it (I tend to prefer squash merging). The only real comment of any substance that I have is that it would be better if the test was a unit test, and then the function would be rewritten slightly to accommodate that (we don't need to test that SBT does the right thing, only that we construct the right command)

jmpage:
@fommil I would much rather a unit test as well; I'm definitely going to try to do that before merging.

I separated this out into two commits since one is doing a little bit of cleanup as a pre-requisite but, otherwise, is unrelated to adding support for ensimeCompileOnly. Would you still prefer me to squash both commits before merging?

fommil:
Github does the squashing now. When you press merge it'll ask if you want to merge or squash. Which means no more need to force push.

ensime-sbt.el
```diff
@@ -144,14 +151,13 @@ again."
...
 
-(defun ensime-sbt-find-subproject (file-name source-set)
-  (when source-set
-    (let* ((config (-> (ensime-connection) ensime-config))
-           (subprojects (plist-get config :subprojects))
-           (matches-subproject-dir? (lambda (dir) (string-match-p dir file-name)))
-           (find-subproject (lambda (sp)
-                              (-any matches-subproject-dir? (plist-get sp :source-roots)))))
-      (-> (-find find-subproject subprojects) (plist-get :name)))))
+(defun ensime-sbt-find-subproject (file-name)
```
fommil: I was reading the emacs lisp book today on the train and I discovered `locate-file`, which might be a much better way to implement this. Also, I think this function is general enough to be used outside of the ensime sbt support, perhaps rename and move to `ensime-config`?

FYI if you want to know where `->` comes from (because the `C-h f` description doesn't do it justice) it's from clojure. https://clojuredocs.org/clojure.core/-%3E (it took me to read a book on clojure to understand what was going on).

ensime-sbt.el
```diff
@@ -144,14 +151,13 @@ again."
...
+  (let* ((config (-> (ensime-connection) ensime-config))
+         (subprojects (plist-get config :subprojects))
+         (matches-subproject-dir? (lambda (dir) (string-match-p dir file-name)))
```
fommil: heh, I think my lisp foo has increased since this code was merged... I would definitely want to see functions defined with `cl-flet` now, not a scheme-style `let`.

jmpage:
@fommil I would like to mess around with this more but might not get around to it until the weekend. Would you prefer me to close this PR until then?

fommil:
Let's leave it open. No rush.

fommil:
I'm merging this because I want to start using it :stuck_out_tongue: but feel free to send follow up PRs with any updates that you want... including breaking changes. We'll consider it an undocumented feature for now. When you're happy with it, add it to http://ensime.github.io/editors/emacs/cheat_sheet/

jmpage:
Sure thing. Thanks for the heads up.

PR #480, opened 2016-06-23
Add autoload cookie for ensime-mode

phillord:
Add a single autoload cookie.

This should mean that installation instructions could be simplfied to:

    (use-package ensime
       :config (add-hook 'scala-mode 'ensime-mode))

Or 

    (use-package ensime
       :ensure t
       :config (add-hook 'scala-mode 'ensime-mode))

and use-package will install it also. Perhaps even better, put the `add-hook' form into ensime-mode, then autoload that. Then installation instructions would be "M-x package-install ensime". I can add this also.

### DIFF
--- a/ensime-mode.el
+++ b/ensime-mode.el
@@ -217,6 +217,7 @@
     ["Troubleshooting" ensime-troubleshooting]
     ))
 
+;;;###autoload
 (define-minor-mode ensime-mode
   "ENSIME: The ENhanced Scala Interaction Mode for Emacs (minor-mode).
 \\{ensime-mode-map}"


fommil:
thanks for the PR, but this won't work. You're supposed to type `M-x ensime`, not `M-x ensime-mode` to start ensime as it requires the server to start up. Did you decided to skip the installation instructions at http://ensime.github.io/editors/emacs/install/#installing :stuck_out_tongue_winking_eye:

fommil:
can you update the PR to autoload `M-x ensime` as well to simplify the installation instructions any further?

phillord:
The installation instructions say:

    (use-package ensime :command (ensime ensime-mode))

there is no necessity for the `:command` section (at least not after my PR). M-x ensime already has an autoload cookie. Did you decide to skip reading the source at https://github.com/ensime/ensime-emacs/blob/master/ensime.el :stuck_out_tongue_winking_eye: ?

fommil:
but why do you need to autoload `ensime-mode`? It is loaded automatically as soon as the user types `M-x ensime` or opens a scala file (thanks to the scala hook).

fommil:
oh right, I see, so this makes both of them auto load.

ok, thanks! If you could update the installation instructions (just click "edit") that'd be appreciated!

fommil:
LGTM

phillord:
Will update the installation instructions tomorrow, once melpa has rebuild. Unless I am too depressed at the election results.

fommil:
you and me both :disappointed:

fommil:
:cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: 




:cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry: :cry:

phillord:
Yep. Going to move back to Scotland, so I can leave the UK with them.

PR #481, opened 2016-06-23
Add hook into scala mode

phillord:
Should have added this to last -- apologies.

This should automatically add the scala-mode-hook during package initialisation. This should shrink the installation instructions to a minimum.


fommil:
heh, yeah that is nice. Cool! This autoload stuff always confused me.

fommil:
LGTM

PR #483, opened 2016-06-25
Server 1.0

### LINKED ISSUE #446: sam's defaults are ensime's defaults
I've got a few defaults, e.g. buffer names and not typechecking on idle, that I'd like to make the defaults for 1.0.

### LINKED ISSUE #456: don't regenerate the classpath when the server version is a release
e.g. server version 1.0.0, don't ever rebuild the classpath.

### LINKED ISSUE #467: update link to old wiki page
should point to the troubleshooting guide

### LINKED ISSUE #470: imenu failing for some payloads
still need to work out what's missing from this payload, but it looks like we could just filter client-side

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  ensime-internalize-offset(nil)
  ensime--flatten-structure-view((:keyword "def" :name "log" :position (:type empty)) nil "MyThing")
```

fommil:
- use 1.0.0 of the server
- close #446 
- clean up the modeline
- close #467
- close #456
- close #470

### DIFF
--- a/ensime-config.el
+++ b/ensime-config.el
@@ -147,10 +147,7 @@ NO-REF-SOURCES allows skipping the extracted dependencies."
         file
       (warn (concat
               "Could not find an ENSIME project file. "
-              "Please see the ENSIME guide: "
-              "https://github.com/ensime/ensime-server/wiki/Quick-Start-Guide "
-              "for instructions on how to write or "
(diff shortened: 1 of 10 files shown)


fommil:
LGTM

PR #487, opened 2016-06-26
Wrap read-char-choice in save-excursion

### LINKED ISSUE #486: Auto-complete parameters inserted inside method name
### Steps to reproduce

1.  Type `List(1, 2, 3).fo`
2. In the autocomplete pop-up select `fold`
3. When prompted for "{ or (" enter {

### Expected behaviour

`List(1, 2, 3).fold(z: A1) { (_0: A1, _1: A1) => A1 }`

### Actual behaviour

`List(1, 2, 3).fo(z: A1) { (_0: A1, _1: A1) => A1 }ld`

marcsaegesser:
Fixes #486.

The call to read-char-choice to prompt for { or ( causes point to be moved back to the last position the user typed a character rather than after the text that auto-completion just wrote to the buffer. The result is that method parameters get insert inside the auto-completed method name.

### DIFF
--- a/ensime-company.el
+++ b/ensime-company.el
@@ -223,7 +223,7 @@ been inserted immediately prior to the point."
                    (equal
                     ?\{
                     (or force-block
-                        (read-char-choice "{ or (" '(?\{ ?\())))))
+                        (save-excursion (read-char-choice "{ or (" '(?\{ ?\()))))))
              (snippet
               (ensime--build-yasnippet-for-call
                param-sections


fommil:
heh, sweet, I was just trying to reproduce when I hit a regression I introduced yesterday.

fommil:
LGTM

fommil:
looks like a transient test failure, I'll kick it off again

fommil:
no test though, so we could get regressions in the future. Any chance of adding a test around this?

marcsaegesser:
I need to figure out how to write an Emacs Lisp unit test that can work with something interactive like read-char-choice. I'm sure it's possible, but I'm still an elisp novice so I have some more learning to do. Pointers to good resources are welcome.

fommil:
we have some links on http://ensime.github.io/editors/emacs/contributing/ a unit test would be best and will involve mocking out the server messages. I'd like to move us to more cucumber style tests for user interactions.

PR #494, opened 2016-07-02
Fix ensime-edit-definition-other-frame not working

jeanny21:
`ensime-edit-definition-other-frame` opened definition of symbol in other window instead of opening in other frame. Adding another predicate related to frame fix the issue.


fommil:
LGTM but this whole feature seems to me designed wrong from the beginning, we shouldn't need to be passing flags around like this.

PR #496, opened 2016-07-06
Move point at beginning of a Startup notification

VlachJosef:
When showing Startup notification we should display beginning rather than end of the buffer. The same holds for Welcome message.


fommil:
LGTM don't know why drone is acting up

fommil:
Oh, because I'm an idiot

```
cask pkg-file: error: Emacs does not exist at /opt/emacs-24.5/bin/emacs.  Did you install Emacs?
```

I'll fix that quickly...

fommil:
ok, fixed now. Also I learnt that our CI runs emacs 24.4, not 24.5.

fommil:
*sigh* the CI is broken.

PR #497, opened 2016-07-07
Remove text properties from search input

### LINKED ISSUE #482: ensime-search causes the server to explode
if one copies and pastes a search into the search bar it'll create a weird marked up property search that isn't valid SWANK and blows up the server.

marcsaegesser:
Remove any text properties (these are preserved for yanked text) before submitting a search query to the server.

The fix for the issue was trivial. The real effort was making the unit test run. There was an existing "Interactive search test" in ensime-test.el that was commented out; probably because it crashed. That problem turned out to be caused by ``split-window`` failing because the search window was too small to be split into the results/entry areas. The tests run in emacs --batch mode which provides a very small frame (about 10x10) which is too small to contain all the windows.

We now gracefully reduce the size of the text entry area based on the size of the search window. I expect no one will ever run into this outside of the unit tests.

Fixes #482

### DIFF
--- a/ensime-search.el
+++ b/ensime-search.el
@@ -161,7 +161,7 @@
 
 
 (defun ensime-search-mode ()
-  "Major mode for incrementally seaching through all open buffers."
+  "Major mode for incrementally searching through all open buffers."
   (interactive)
   (setq major-mode 'ensime-search-mode
         mode-name "ensime-search")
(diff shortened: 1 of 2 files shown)


fommil:
LGTM this is really cool! It's always bothered me

marcsaegesser:
I'm ashamed to admit how long I spent figuring out the unit test problem. On the plus side I'm having a lot of fun getting into Emacs Lisp.

Issue #383, opened 2016-02-29
Leaving paste mode kills the repl session


xificurC:
### Steps to reproduce
1. Open a scala file and start ensime (`M-x ensime`)
2. Start the repl (`C-c C-v z`)
3. Enter paste mode (type ":paste" into the repl)
4. Leave paste mode with `C-d`
### Expected behaviour

Paste mode successfully finished
### Actual behaviour

Paste mode successfully finished, repl killed right afterwards.
### What is paste mode for

Paste mode is e.g. needed when defining a class with a companion object. In a normal file one could have

``` scala
class Something {
  ...
}

object Something {
  ...
}
```

and compiling that woks just fine. Entering it into the repl will not work correctly though because the class would be defined with the closing bracket and once that's done we cannot implement a companion object. That's why there's paste mode, the repl will read everything in 1 bulk and then this works.

fommil:
can you give some more details in the reproduction please? People might not know what paste mode is, or which exact ensime-emacs commands to run to get there.

Some example scala code to paste would also be good that demonstrates the need to `:paste`

xificurC:
Added more info

fommil:
we're been chatting on gitter, do you feel you have enough information about how this works to try and put together a fix? We're happy to help (although I'm not the greatest elisper in the world)

fommil:
@xificurC ! Ping

xificurC:
Hey @fommil , sorry but I'm low on free time so I'm not going to chase this

PR #502, opened 2016-07-20
Send regions to the inferior REPL in paste mode.

ieure:
Thank you for contributing to ENSIME!

If this is your first time submitting a PR, this is what you can expect:

1. You'll have to agree to the licences we use. You'll get a link to follow in a moment.
2. Your code will be automatically tested. That can take as long as an hour before the results come back. You can read `.drone.yml` to see what is being run.
3. We'll try to review your code and give feedback quickly.
4. Please squash your commits into independent units. We use the commit log as our changelog so it is good if your commit messages are clear and concise.
5. You can request an ENSIME sticker from `@fommil`, be sure to ask if you want one!
6. If everything goes silent, please ping us.

Now, delete this line and everything above it, then tell us which tickets this PR closes (please don't reference tickets directly from your commit messages).

If you send a piece of code containing forward references, simply
dumping it into the REPL doesn’t work because, it processes each
definition individually. Enabling paste mode, this allows the REPL to
correctly deal with this situation.

### DIFF
--- a/ensime-inf.el
+++ b/ensime-inf.el
@@ -234,8 +234,13 @@ the current project's dependencies. Returns list of form (cmd [arg]*)"
   "Send current region to Scala interpreter."
   (interactive "r")
   (ensime-inf-assert-running)
-  (comint-send-region ensime-inf-buffer-name start end)
-  (comint-send-string ensime-inf-buffer-name "\n"))
+
+  (let ((reg (buffer-substring-no-properties start end)))
+    (with-current-buffer ensime-inf-buffer-name
+      (comint-send-string nil ":paste\n")
+      (comint-send-string nil reg)
+      (comint-send-string nil "\n")
+      (comint-send-eof))))
 
 (defun ensime-inf-eval-definition ()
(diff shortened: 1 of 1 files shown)


fommil:
Cool. Gold you come up with an automated test for this please?

fommil:
#383 might need fixing for this to be useable

ieure:
> Cool. Could you please come up with an automated test for this?

I don't think the effort is warranted, any test would be both heavy (due to needing to create a process and buffer to send a region of) and brittle, due to duplicating code in the actual function.

> #383 might need fixing for this to be useable

I can't reproduce this issue. Both manually entering paste mode and my changes in this PR work fine.

fommil:
@ieure if it doesn't have a test, then you can't expect it to keep working and we put a lot of effort into testing so I disagree with your diagnosis. This is documented in the contributing page if you'd like to learn more.

I'll leave this PR here until you have a think about testing. We're a mature project now and testing is important to us.

ieure:
Okay, never mind then.

PR #504, opened 2016-07-31
flush blank lines when sending region to Scala interpreter

lxol:
When using `ensime-inf-eval-region` to evaluate a Scala expression with multiple blank lines, the interpreter produces the following error:
````
scala>      |      | You typed two blank lines.  Starting a new command.

scala> <console>:1: error: eof expected but '}' found.
}

`````
The fix is to remove all blank lines from the region before sending it to `ensime-inf-eval-region`.

### DIFF
--- a/ensime-inf.el
+++ b/ensime-inf.el
@@ -242,6 +242,19 @@ the current project's dependencies. Returns list of form (cmd [arg]*)"
       (comint-send-string nil "\n")
       (comint-send-eof))))
 
+(defun ensime-inf-eval-result ()
+  "Get REPL evaluation result."
+  (with-current-buffer ensime-inf-buffer-name
+    (save-excursion
+      (goto-char (point-max))
+      (next-line -1)
+      (end-of-line)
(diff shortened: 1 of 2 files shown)


fommil:
do you think you can come up with a test for this?

fommil:
#502 seemed to be doing the same thing, but I am nervous about merging untested code.

lxol:
I'll try to come up with the test. May be paste mode #502 is even better solution for the issue.

fommil:
Tests would be good and will be really necessary for the repl cider stuff you were looking into

lxol:
I've replaced flush lines change with  @ieure paste mode and added a test.

fommil:
great stuff! Are you coming to the next hack day? I know a few people who are interested in a general framework for evaluation/display like cider now

fommil:
LGTM

lxol:
yes, I am planning to come.  I thought @kozikow was very close to implementing the cider type evaluation.

PR #507, opened 2016-08-11
Fix sbt-do-test-only when the cursor is inside a string

chessman:
Steps to reproduce:
1. Open RichPresentationCompilerSpec.scala.
2. Place the cursor on any string inside RichPresentationCompilerSpec.
3. M-x ensime-sbt-do-test-only-dwim.

It runs RichPresentationCompilerThatNeedsJavaLibsSpec instead.

The problem is in incorrect behavior of backward-up-list when the cursor is inside a string. The workaround is ensuring that backward-up-list is not called from a string.

"inside-string?" predicate was taken from here: http://ergoemacs.org/emacs/elisp_determine_cursor_inside_string_or_comment.html


fommil:
Lgtm

PR #512, opened 2016-08-18
evaluate in REPL and display inline like CIDER

lxol:


fommil:
:-O

mmlac:
Hi @lxol, I am trying to find the function `ensime-inspect-last-result` that you are referencing in `ensime-overlay.el L151`, and I can't seem to find out where it is defined. Is that still a todo or is it in another package?

PR #515, opened 2016-08-25
option to display type and errors at point inline

lxol:
ensime-type-at-point, ensime-type-at-point-full-name and
ensime-print-errors-at-point with "C-u C-u" prefix argument display
the result inline using the CIDER style overlay.


fommil:
LGTM looking forward to trying this out!

PR #516, opened 2016-08-30
default ensime server version to stable 1.0.0

mpollmeier:
wouldn't it make more sense to default ensime-server to a stable version? as it stands, every user gets a warning about using an unstable snapshot, and then needs to either configure the stable version or make a configuration to acknowledge that they are using bleeding edge...

### DIFF
--- a/ensime-vars.el
+++ b/ensime-vars.el
@@ -70,7 +70,7 @@ works for sbt projects."
   :group 'ensime-ui)
 
 (defcustom ensime-server-version
-  (or (getenv "ENSIME_SERVER_VERSION") "2.0.0-SNAPSHOT")
+  (or (getenv "ENSIME_SERVER_VERSION") "1.0.0")
   "Distributed version of the server to upgrade and start.
 This is primarily useful for ENSIME developers (or bug reporters)
 to test against. The client is designed to work with the default


fommil:
@mpollmeier somebody hasn't read the installation instructions[1] or the branching strategy[2], have they? ;-)

- [1] http://ensime.github.io/editors/emacs/install/#stable
- [2] https://github.com/ensime/ensime-emacs/tree/1.0

fommil:
Btw it is absolutely 100% intended that users acknowledge that they are using bleeding edge. I have no intentions of changing this.

fommil:
and you'll also find that the `master` branch is incompatible with server `1.0.0` in several places, and about to be in a **lot** of places.

In keeping with the ENSIME philosophy of "ENSIME appeals to hackers, minimalists and connoisseurs - engineers and artists who craft their own exquisite tools and help their neighbour" you either roll up your sleeves and help when it breaks, or you drop down to the "stable" release where nothing is going to change.

mpollmeier:
True, I never read those - I use [spacemacs](https://github.com/syl20bnr/spacemacs/) which installs ensime for me in it's [scala layer](https://github.com/syl20bnr/spacemacs/blob/master/layers/%2Blang/scala/packages.el). 

I agree it's a good idea for users to acknowledge if they're using bleeding edge. In respect of the hacker's philosophy there, maybe it makes sense to amend the warning text when using SNAPSHOT as it sounds rather discouraging to do so? 

```
If you did not intentionally do this, it means that you are using
the unstable (developer) release of ensime (e.g. from MELPA),
which is *not recommended*.
```

fommil:
Yeah, that doesn't sound very encouraging does it. Can't you send some suggested text instead in a PR? Also I thought spacemacs fixed this to use stable by default.

chrisbarrett:
If I understand correctly, Spacemacs doesn't have a way to use packages from MELPA Stable yet. They're tracking this in syl20bnr/spacemacs#6531.

PR #519, opened 2016-09-05
Remove needless argument

syohex:
major-mode functions does not take any arguments.

### DIFF
--- a/ensime-mode.el
+++ b/ensime-mode.el
@@ -141,12 +141,6 @@
   "Keymap for ENSIME mode."
   )
 
-;;;;; ensime-mode
-(defun ensime-scala-mode-hook ()
-  "Convenient hook function to start `ensime-mode'."
-  (scala-mode 1)
-  (ensime-mode 1))
-
 (defun ensime-run-after-save-hooks ()
   "Things to run whenever a source buffer is saved."
   (when (ensime-source-file-p)


syohex:
BTW what usecase is this function(`ensime-scala-mode-hook`) ? If this function is set `scala-mode-hook`, then it causes infinite loop(`scala-mode-hook` calls `ensime-scala-mode-hook`, `ensime-scala-mode-hook` calls `scala-mode`, `scala-mode` calls `run-hook 'scala-mode-hook`, .....).

`el-get` recipe sets `ensime-scala-mode-hook` to `scala-mode-hook` :-(
- https://github.com/dimitri/el-get/blob/f355aef81d620f2477875995340b209ff8c5e06d/recipes/ensime.rcp#L15

fommil:
I'd prefer if you just deleted this function. Can you please update your PR to do that?

There seems to be a bug in the `el-get` recipe. Please send a PR to them to fix their recipe, whoever maintains it can ask questions here or on the gitter channel. The correct startup mechanism should be to autoload everything that is declared with an autoload cookie. We have this lazy autoload:

```elisp
;;;###autoload
(add-hook 'scala-mode-hook
          (lambda () (when (boundp 'ensime) (ensime-mode))))
```

which means that `scala-mode` will load `ensime-mode` (not the other way round) only if the `ensime` function is bound. This allows the user to define `ensime` lazily and have fast console sessions just using `scala-mode`.

fommil:
@npostavs can you please update the recipe accordingly? some instructions above.

syohex:
> I'd prefer if you just deleted this function. Can you please update your PR to do that?

Will do.

syohex:
I have updated.

fommil:
this will probably break things for people who are tracking the dev release, so be prepared for some noise :smile: thanks!

PR #520, opened 2016-09-05
Launch from cask

phillord:
Just a starter for 10 -- it doesn't fix all the problems, some of the tests do not succeed, maybe some do not run, but at least you don't have to set the path up.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+cask emacs --batch -q --directory=. --load test/ensime-emacs-test.el --funcall ert-run-tests-batch


fommil:
thanks! The unit tests are being launched via cask, so they are ok... it's the integration tests that are using their own start script. This isn't really a PR that can be accepted since it doesn't do anything, but it's good to have it around.

phillord:
These are the ones in ensime-test.el? Same issue there, perhaps? If you change `exec emacs` to `exec cask`, then there is at least a vague chance that it will all work.

fommil:
If you want to tackle it all the way through that'd be ace! Will cask create all the autoload files? That's the main reason I've used package.

phillord:
cask can create autoload files yes, and the package descriptor file.

fommil:
yeah, we created the package descriptor already but I couldn't find docs on the autoload, so I build the package and then used `use-package` to ensure the autoloads were built correctly for testing.

fommil:
there is no real reason why we need the shell script if we could launch the tests via cask and it will probably simplify the dot file.

PR #524, opened 2016-09-21
Support Play's default dir layout

z4139jq:
Play does not follow the SBT standard dir layout, it is sweet to support that though it is configurable to re-layout


fommil:
This will all be deprecated soon when we finish 2.0 support and upgrade the .ensime format. For now, LGTM

PR #525, opened 2016-10-10
Fix syntax error in build.sbt

mattiasb:


mattiasb:
Basically I was getting this whenever i ran (ensime-server-update):
```
/tmp/ensime_update_87764qG/build.sbt:7: error: eof expected but ';' found.
scalaBinaryVersion := "2.10"
^
[error] Error parsing expression.  Ensure that settings are separated by blank lines.
Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? q
```

Adding a newline at the right point seems to fix it.

mattiasb:
```
$ sbt --version
sbt launcher version 0.13.1
```

fommil:
We don't support this old version of sbt and besides it should be picking up the build.properties file to use the version we specify. Upgrade your sbt.

fommil:
If you can't upgrade sbt, then use assembly jars as documented on the contributing page.

PR #527, opened 2016-10-15
Set ensime-inf-overlay-marker in sbt buffers

greenrd:


fommil:
@greenrd this introduced a regression for me when I ran `util/testOnly *EnsimeFileSpec` in my dev branch of `ensime-server`. Should the `ensime-sbt` perhaps depend on some package to ensure that this is defined first? Or you could wrap the `nil` with a guard.

```
Debugger entered--Lisp error: (void-variable ensime-inf-overlay-marker)
  ensime-inf-postoutput-filter("[error] Total time: 3 s, completed 15-Oct-2016 18:23:23\n> ")
  run-hook-with-args(ensime-inf-postoutput-filter "[error] Total time: 3 s, completed 15-Oct-2016 18:23:23\n> ")
  comint-output-filter(#<process *sbt*<~/Projects/ensime-server/>> "[error] Total time: 3 s, completed 15-Oct-2016 18:23:23\n> ")
```

greenrd:
@fommil I can't reproduce that. And that looks like a manifestation of the _original bug_, not my fix. Are you sure you loaded my fix?

fommil:
I'll try again. If this is the original bug then the fix is wrong, it means a require statement is missing.

greenrd:
@fommil Did you get a chance to try this again?

fommil:
after reverting your commit, I haven't seen it anymore. Can you try creating a reproduction for the original bug you seen?

greenrd:
But `setq` doesn't need a `boundp` check, and `void-variable` doesn't mean a variable is `nil`, it means it's unset, so you must be confused. All I am doing is `setq` and if you try executing `(setq somethingxyz nil)` in your scratch buffer it will not error. Also the stack trace does not refer to my code.

fommil:
it is just weird to be `setq`ing something that seems completely irrelevant, it feels like a band aid rather than addressing the core problem. Until we understand the cause of the stacktrace you seen, we can't really address it... can you please put together a reproduction?

PR #533, opened 2016-10-21
make the submodule lookup more general, and Windows compatible

### LINKED ISSUE #530: (convert-standard-filename buffer-file-name)
in order to do pathname matching and so on, we should be using

```
(convert-standard-filename buffer-file-name)
```

to make sure `\` is preserved in windows file names, e.g. to compare against the dirs from the `.ensime`

fommil:
close #530 and introduce breaking changes for @yyadavalli  (sorry!)

@jmpage if you're around, do you mind doing a review please?

### DIFF
--- a/ensime-config.el
+++ b/ensime-config.el
@@ -34,12 +34,27 @@
   (let ((connection (ensime-connection)))
     (ensime-config connection)))
 
-(defun ensime-process-for-config (config)
-  "Obtain the ENSIME Server process for the given config."
+(defun ensime-process-for-config (&optional config)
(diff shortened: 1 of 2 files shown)


fommil:
we should really make all these `-for-config` functions take an `&optional config` and auto detect if it is not supplied.

fommil:
I should really have put windows tests in place... but the amount of time it would take me would be best spend just rewriting the bits of ensime that are likely to break on windows!

fommil:
I broke all the things.

Issue #531, opened 2016-10-21
Use ensime-sbt's ensimeScalariformTask for ensime-format-source


yyadavalli:
https://github.com/ensime/ensime-sbt/pull/230

fommil:
FYI #530

fommil:
btw, I'm trialing this (I added a `save-buffer`)

```elisp
(defun ensime-sbt-do-scalariform-only ()
  (interactive)
  (save-buffer)
  (let ((subproject (ensime-sbt-find-subproject buffer-file-name)))
    (if subproject
        (sbt-command (concat subproject "/ensimeScalariformOnly " buffer-file-name))
      (sbt-command (concat "ensimeScalariformOnly " buffer-file-name)))))
```

the concept of running a command for a specific subproject seems like something we could factor out.

fommil:
sounds good!

fommil:
we'll give the other editors a heads up before removing it from the server, and then remove it anyway MUAHAHAHAHAHA!

fommil:
sorry, did I say that out loud?

yyadavalli:
So, should I remove the command or just call `ensime-sbt-do-scalariform-only`?
I'd say remove it.

fommil:
remove! I was talking about server side :-) sorry I broke the functions you're calling with my latest PR. Should be a trivial update for you though.

yyadavalli:
Sure, I believe no other editors are using those messages. If we were to remove `formattingPreferences` from `EnsimeConfig` that would have more impact.

PR #534, opened 2016-10-22
Use ensimeScalariformOnly to format buffer

yyadavalli:
- Add ensime-sbt-do-scalariform-only functions
- Refactor common code with ensime-sbt-do-compile-only


yyadavalli:
Resolve #531

fommil:
looks like you've not caught all references to `ensime-format-source`

yyadavalli:
Yup, I think I missed the keybinding and the test. I am trying to get the test working.

PR #540, opened 2016-10-24
Create ensime-inf-overlay-marker variable correctly

### LINKED ISSUE #526: void-variable ensime-inf-overlay-marker in sbt buffer
### Steps to reproduce

With unstable version:

1. `M-x ensime`
2. `C-c C-b s`
3. Enable Enter Debugger on Error in emacs options
4. press Enter

### Expected behaviour
No error

### Actual behaviour
Error:
```
Debugger entered--Lisp error: (void-variable ensime-inf-overlay-marker)
  ensime-inf-postoutput-filter("> ")
  run-hook-with-args(ensime-inf-postoutput-filter "> ")
  comint-output-filter(#<process *sbt*<~/git/drivetribe/backend/>> "> ")
  accept-process-output(#<process *sbt*<~/git/drivetribe/backend/>>)
  comint-send-input()
  funcall-interactively(comint-send-input)
  call-interactively(comint-send-input nil nil)
  command-execute(comint-send-input)
```

greenrd:
This ensures it will only be set to nil if not already set.

Fixes #526 in the correct way.

### DIFF
--- a/ensime-inf.el
+++ b/ensime-inf.el
@@ -81,6 +81,8 @@ with data loaded from server."
 Caches the last pair used in the last ensime-inf-load-file.
 Used for determining the default in the next one.")
 
+(defvar ensime-inf-overlay-marker nil)
+
 
 (define-derived-mode ensime-inf-mode comint-mode "Scala REPL"
   "Major mode for interacting with a Scala interpreter."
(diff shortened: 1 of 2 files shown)


fommil:
Cool, thanks for looking into it further.

PR #541, opened 2016-10-25
Make sbt tasks work with test files

### LINKED ISSUE #539: formatting doesn't work for test directories
@yyadavalli i tried formatting a test file and it ran `myproject/ensimeScalariformOnly` which collapses to `myproject/compile:ensimeScalariformOnly`.

Discovering the phase (compile/test/it/etc) is difficult right now, perhaps we should update the sbt plugin to have the task on the `Project` instead of the config?

yyadavalli:
Simple but not so elegant fix to make `ensimeCompileOnly` and `ensimeScalariformOnly` work with files in test folders

Close #539

### DIFF
--- a/ensime-sbt.el
+++ b/ensime-sbt.el
@@ -71,8 +71,12 @@
   "Run a sbt COMMAND in the module containing FILE-NAME, if specified."
   (let ((subproject (ensime-subproject-for-config)))
     (if subproject
-        (sbt-command (concat subproject "/" command " " file-name))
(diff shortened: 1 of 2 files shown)


fommil:
Tests are broken

fommil:
Just remove the compileOnly test. The interplay with sbt is too unreliable. We'll unit test it once the .ensime format is updated.

fommil:
thanks again!

yyadavalli:
welcome..

PR #546, opened 2016-10-30
ensure default-directory ends with slash

eikek:
Hello,
generating the ensime config file via the `ensime-refresh-config` (from lisp, not interactively) didn't work for me. It created the config file in the wrong folder (not in the project directory but one above). I then found that the `default-directory` used for the sbt process didn't end in a slash, and changing that fixed my problem. But it works when calling it interactively from within a scala buffer that is connected to ensime (i.e. there is a sbt project for this scala buffer and ensime is connected). I don't know why that is…. I didn't investigate further, because I thought it would be better to always wrap the dirname in `file-name-as-directory` when setting it to `default-directory`?

### DIFF
--- a/ensime-config.el
+++ b/ensime-config.el
@@ -218,7 +218,7 @@ project directories, because neither does ensime-sbt.)"
 (defun ensime--refresh-config-sbt (project-root task on-success-fn)
   (with-current-buffer (get-buffer-create "*ensime-gen-config*")
     (erase-buffer)
-      (let ((default-directory project-root))
+      (let ((default-directory (file-name-as-directory project-root)))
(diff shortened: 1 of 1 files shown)


fommil:
We should probably just delete this functionality to be honest, or move to an "ensime-contrib". The format has changed and this file isn't supported by the server anymore.

eikek:
Ok, then it's probably not worth merging this. Otherwise if it's moved ensime-contrib, it may be useful again.

fommil:
doesn't hurt to merge for now, thanks!

eikek:
Thanks!

Issue #542, opened 2016-10-25
M-. in indirect buffer kills Ensime


VlachJosef:
### Steps to reproduce
1. Create file `Main.scala`:

``` scala
package core

object Main {
  def one: String = two
  def two: String = "2"
}
```
1. Run `ensime` command
2. Clone `Main.scala` buffer using `clone-indirect-buffer-other-window` (`C-x 4 c`)
3. Navigate to definition of a method `two` in indirect (cloned) buffer. (ie. `M-.` on line  `def one: String = two`)
### Expected behaviour

Jump to definition of `two` on line `def two: String = "2"`
### Actual behaviour

Ensime is non responsive from now on.

```
16:38:30.157 ERROR akka://ENSIME/user/ensime-main/tcp-server/con1 o.e.s.t.TCPConnectionActor - Error seen during message processing, closing client connection
org.ensime.server.protocol.swank.SwankRPCFormatException: Invalid rpc request (swank:symbol-at-point nil 596)
    at org.ensime.server.protocol.swank.SwankProtocolRequest$RpcRequestEnvelopeFormat$.read(SwankFormats.scala:860)
    at org.ensime.server.protocol.swank.SwankProtocolRequest$RpcRequestEnvelopeFormat$.read(SwankFormats.scala:849)
    at org.ensime.sexp.Sexp.convertTo(Sexp.scala:21)
    at org.ensime.server.SwankProtocol.decode(SwankProtocol.scala:17)
    at org.ensime.server.tcp.TCPConnectionActor.repeatedDecode(TCPConnectionActor.scala:108)
    at org.ensime.server.tcp.TCPConnectionActor.attemptProcess(TCPConnectionActor.scala:97)
    at org.ensime.server.tcp.TCPConnectionActor$$anonfun$incoming$1.applyOrElse(TCPConnectionActor.scala:49)
    at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:170)
    at akka.actor.Actor$class.aroundReceive(Actor.scala:467)
    at org.ensime.server.tcp.TCPConnectionActor.aroundReceive(TCPConnectionActor.scala:17)
    at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
    at akka.actor.ActorCell.invoke(ActorCell.scala:487)
    at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:238)
    at akka.dispatch.Mailbox.run(Mailbox.scala:220)
[...]

PR #551, opened 2016-11-13
M-. in indirect buffer kills Ensime (#542)

VlachJosef:
It turns out that almost any operation in indirect buffers caused ensime to crash. This PR is fixing it, so now it is possible to do any operation in indirect buffer as if it was regular buffer.


fommil:
sorry! I'm taking forever. Looks like an unstable test, fixed in another PR

fommil:
living the wild life, merging red builds

PR #563, opened 2016-11-30
Dotensime launcher

fommil:
this is really seriously hacky and needs a refactor because half the functions now have names that make no sense, and the other half can probably be deleted wholesale.

@raulraja this, in combination with the latest sbt-ensime, should solve the typelevel scala jar usage problem. Would you like to be our QA tester and give this a try?


fommil:
hmm, given that the tests are hanging... that might be quite brave.

raulraja:
Sam, I'll be traveling to London tomorrow but as soon as I'm close to a
computer I'll give it a shot. What cached files do I need to remove to
ensure it's not fetching for local cached repositories that ensime uses?

fommil:
not sure I follow what you mean. You might need to clear out your coursier cache because of various conflicting cache formats, but you should probably have done that already by now.

fommil:
FYI to QA install this, check out my branch and run `cask package` (assuming you have cask installed). Then, `M-x package-delete RET ensime` followed by  `M-x package-install-file RET <your checkout>/dist/ensime-2.0.0.jar`

fommil:
hmm, this seems to be passing the tests ... so I'm tempted to just merge it.

fommil:
oh, screw it... the worst that could happen is that somebody sends a PR to fix it and everybody just rolls back to the stable release.

This'll almost certainly catch everybody who hasn't upgraded `sbt-ensime` in a long time.

raulraja:
Awesome! Looking forward to try it!

fommil:
make sure to read the sbt plugin installation instructions again, i.e. don't forget about `ensimeServerVersion`

Issue #523, opened 2016-09-13
import-at-point transformation (from _._ to _.{_, _}) adds extra i


Trevoke:
### Steps to reproduce
1. Have a file with `import java.util.GregorianCalendar`
2. Write `Calendar`
3. With `Calendar` on point, call ensime's function to import type at point
### Expected behaviour

Should display `import java.util.{ Calendar, GregorianCalendar }`
### Actual behaviour

Displays `iimport java.util.{Calendar, GregorianCalendar }`

fommil:
can you minimise this as a starting file or project? I can't reproduce.

Trevoke:
=> https://github.com/Trevoke/for-ensime-debugging-purposes

fommil:
+1 no package definition, we must be assuming there is one.

PR #570, opened 2016-12-21
Fixed inserting in import groups at point-min

bomgar:
Should fix https://github.com/ensime/ensime-emacs/issues/523

The importing code is really hard to get right. Isn't this something the server could do?

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -602,7 +602,10 @@ Returns a function/closure to invoke the necessary buffer operations to perform 
                            (s-split ",") (-map 's-trim)
                            (cons qualified-class-name) (-sort 's-less?) (s-join ", "))))
     (lambda ()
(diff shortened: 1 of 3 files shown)


fommil:
yeah I agree... there might be a ticket there. would you be interested in hacking on that on the server side? We have a pretty clear refactoring api and client support.

bomgar:
I'm interested in doing that but I'm not sure if I have enough time. I might give it a try.

fommil:
https://github.com/ensime/ensime-server/issues/1401 marked as externally blocked, no reason we can't do it ourselves

bomgar:
In the meantime this fix should still be good. I'm not sure about the CI build but I don't think it has something to do with my changes.

fommil:
Great thanks! The build machines are now store again but sadly our Emacs tests are just really flakey.

PR #574, opened 2017-01-01
Eldoc support

yyadavalli:
Use eldoc to display the current symbol and type at point.

Liked the idea while working with `haskell-mode`

WIP: Will add tests, once I get more comments on idea and implementation

### DIFF
--- a/ensime-model.el
+++ b/ensime-model.el
@@ -30,12 +30,6 @@
 (defun ensime-symbol-decl-pos (sym)
   (plist-get sym :decl-pos))
 
-(defun ensime-symbol-name (sym)
-  (plist-get sym :name))
-
-(defun ensime-symbol-local-name (sym)
-  (plist-get sym :local-name))
-
 (defun ensime-symbol-type (sym)
   (plist-get sym :type))
 
--- a/ensime-model.el
+++ b/ensime-model.el
@@ -30,6 +30,12 @@
 (defun ensime-symbol-decl-pos (sym)
   (plist-get sym :decl-pos))
 
+(defun ensime-symbol-name (sym)
+  (plist-get sym :name))
+
+(defun ensime-symbol-local-name (sym)
+  (plist-get sym :local-name))
+
 (defun ensime-symbol-type (sym)
   (plist-get sym :type))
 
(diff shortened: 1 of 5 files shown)


fommil:
oh wow, this could be cool! Did you know we have a popup backend already? Can't remember which one, but yeah this sounds great.

Sorry the CI is broken. Let me know when you're happy with this PR and I'll try it out. I won't be doing much scala for the next few days.

yyadavalli:
I do know we have a popup backend, but I have slightly different use case in mind.

I am trying to do it automatically when browsing the file, without the need to invoke
`ensime-type-at-point`.

The PR is good enough to try, it just lacks tests

fommil:
Sorry I've not found the time to try this out yet

yyadavalli:
No problem, any thoughts on how something like this should be tested.

manuel-uberti:
Any update on this one?

yyadavalli:
Still the tests are missing, but you should be able to use it by manually pulling in my changes.

ensime-model.el
```diff
@@ -30,6 +30,12 @@
 (defun ensime-symbol-decl-pos (sym)
   (plist-get sym :decl-pos))
 
+(defun ensime-symbol-name (sym)
```
martinmcnulty: Is `ensime-symbol-name` used anywhere?

fommil: good catch

ensime-model.el
```diff
@@ -30,6 +30,12 @@
...
+  (plist-get sym :name))
+
+(defun ensime-symbol-local-name (sym)
```
fommil: I'm leaning against these little helpers, and instead preferring common-lisp style destructuring of the contents.

martinmcnulty:
I really like this idea - I miss the ease of access to type info that I used to get from hovers in Eclipse.

For me the only snag at the moment is that I sometimes don't get type info about the symbol at point because there is an implicit conversion happening somewhere else on the same line. I think this is because `(ensime-implicit-notes-at (point))` seems to give you notes on any implicit on the same _line_ as point.  

Might be better to always show the type info (when possible) and put any info about implicits afterwards?

yyadavalli:
@martinmcnulty That would be better, are you interested in taking over this PR and adding that?

I am not sure when I would get time to do it.

yyadavalli:
Hi, @fommil this can be merged, if you are satisfied...

PR #577, opened 2017-01-21
Put .ensime_cache/dep-src/source-jars directory into cache even when

VlachJosef:
it doesn't exists.

This is fix for the case when sources in
.ensime_cache/dep-src/source-jars weren't navigable due to cache
ensime--cache-source-root-set not including
.ensime_cache/dep-src/source-jars since it doesn't existed when the
project were created.


fommil:
INSTAMERGE AWESOMESAUCE!!1!

PR #587, opened 2017-03-18
ensime-inf-eval-region - send current line to Scala interpreter

VlachJosef:
if no region is active send current line to Scala interpreter.


fommil:
melpa stable is down today :cry:

PR #595, opened 2017-04-21
Keybinding and deserialization fixes for expand match cases

iravid:


fommil:
sorry about shitty CI. I've wasted too much time of my life keeping it ticking and I've just decided to let it fail... if I'm going to invest that time into ensime-emacs I'd rather just rewrite it to not need to talk to the server in tests.

Issue #581, opened 2017-02-12
imports should go to top of file, not inner package scope


fommil:
```scala
package foo

import stuff._

package object bar {
   // imports are appearing here, not up above as expected

  Thing.foo // Thing needs an import
}
```

PR #596, opened 2017-04-24
Change logic of searching for imports when importing type at point

nigredo-tori:
Use the first `package` match in a file rather then the last before cursor.


nigredo-tori:
This is a fix for #581

fommil:
Thanks!

fommil:
oh damn, I just remembered why it wasn't this easy.

Consider files that look like this:


```scala
package foo
package bar
package thing
```

we end up putting the import in the wrong place.

PR #598, opened 2017-04-24
bump heap for the graph

fommil:
I have no real hopes of this fixing the tests, but here goes...


fommil:
at least it didn't time out

PR #600, opened 2017-04-25
Better import statement injection logic

nigredo-tori:
Another attempt at fixing #581.
This improves on #596 by handling the case with multiple package statements:
```scala
package foo
package bar

// Import goes here

package object baz
```


fommil:
nice! I would usually request a test, but since our CI is completely broken, I'll let you off.

PR #601, opened 2017-04-27
Add ensime-sbt-send-eol command

ashawley:
- Bind to `C-c C-b C-j` to ENSIME minor mode


fommil:
I'm really not keen on expanding `ensime-sbt` but since there is no alternative...

Issue #186, opened 2015-07-03
Search integration with ido


fommil:
or other typical emacs search interaction tools. Having our own front end gives us some options that we might not get --- e.g. choose source or docs. But I think the vast majority of people just want to jump to source in search. If people want search for documentation we can have a separate function for that where the callback is to ask for docs instead.
##

tbje:
Helm does give you this option, one default one and in addition a multiple choice menu.

fommil:
It would be good to integrate with helm if people want it, but I personally can't stand it so it would have to be optional :-P

fommil:
we have helm support for search now but leaving this open for `ido`.

PR #604, opened 2017-05-03
Ivy integration for ensime search

yyadavalli:
Related to #186 

Basic support for now (goto source only).

Suggestions on more options to add are welcome.

### DIFF
--- a/ensime-search.el
+++ b/ensime-search.el
@@ -24,6 +24,7 @@
   (require 'ensime-macros))
 
 (require 'dash)
+(require 'ensime-helm)
 (require 'ensime-ivy)
 
 (defvar ensime-search-interface)
@@ -74,7 +75,6 @@
 (defvar ensime-search-max-results 50
   "The max number of results to return per rpc call.")
 
-
 (defvar ensime-search-mode-map
   (let ((map (make-sparse-keymap)))
     (define-key map "\C-c\C-q" 'ensime-search-quit)
@@ -104,7 +104,6 @@
 (defvar ensime-search-selection-overlay nil
   "Overlay that highlights the currently selected search result.")
 
-
 (defstruct ensime-search-result
   "A ensime-search search result.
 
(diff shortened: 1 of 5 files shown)


fommil:
Can we make the ivy dependency optional? Like we do for helm.

ensime-search.el
```diff
@@ -24,6 +24,10 @@
   (require 'ensime-macros))
 
 (require 'dash)
+(require 'ensime-ivy)
```
fommil: this needs to be opt-in

ensime-vars.el
```diff
@@ -314,9 +314,10 @@ Possible types: `organizeImport', `rename', `extractLocal',
   :type 'boolean
   :group 'ensime-ui)
 
-(defcustom ensime-use-helm nil
-  "Enable helm for some ensime features"
-  :type 'boolean
+(defcustom ensime-search-interface 'classic
```
fommil: cool, but will breaks things for some people so we need to tweet / gitter about it.

yyadavalli:
I believe, it's optional. I did not add it to the package file.

fommil:
You can't require it if it's optional... that'll just blow up if the user doesn't have it installed. For helm we require the user to require the ensime-helm package but we could probably be smarter by autoloading it if they've set the appropriate mode.

yyadavalli:
See, if this does the trick. 

This is a breaking change only for `helm` users, an announcement on gitter and a tweet should do the trick.

I can cleanup `ensime-helm.el` and `ensime-search.el`, if you don't mind merging unrelated changes.

fommil:
cool, can you confirm that this works and you don't have helm installed?

yyadavalli:
Yeah, that's how I tested it

bomgar:
The ensime-use-helm variable is still used here https://github.com/ensime/ensime-emacs/blob/master/ensime-editor.el#L729

yyadavalli:
I will check, thanks

PR #606, opened 2017-05-03
Provide hint message for dwim commands

ashawley:
Suggest that the dwim test commands can be run in a test source
file to avoid getting prompted for inputs.

Here's an example of one of the messages when a dwim command is
not run from a test source file:

    "Run ‘C-c C-b o’ from a test file to avoid the prompts"

### DIFF
--- a/ensime-sbt.el
+++ b/ensime-sbt.el
@@ -185,7 +185,10 @@ again."
         (-> (ensime-subproject-for-config)
             (concat "/" source-set command)
             sbt:command)
-      (-> (ensime-sbt-prompt-for-test) sbt:command))))
+      (-> (ensime-sbt-prompt-for-test) sbt:command)
+      (message "Run `%s' from a test file to avoid the prompts"
+               (key-description
+                (where-is-internal this-command overriding-local-map t))))))
 
 (defun ensime-sbt-do-test-dwim ()
   (interactive)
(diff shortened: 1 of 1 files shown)


fommil:
I'm not sure about this... it can be a bit annoying and is easily fixed if the user reads the docs.

ashawley:
I couldn't find it in the doc.  Those commands don't have docstrings, either.  However, maybe I wasn't looking in the right place.

Regardless, my point isn't about the documentation, anyway.  Giving users suggestions on what they should have done is just good Emacs etiquette, along the same lines as ENSIME itself (`ensime-startup-notifications`, `ensime-troubleshooting`, `ensime-goto-impl', the compatibility check for`ensime-server-jars`).

fommil:
maybe we should add good docstrings instead?

fommil:
(I tend to make it noisy when people keep asking about it on gitter/twitter/SO... but I wish we didn't have to do that)

ashawley:
> maybe we should add good docstrings instead?

Why not all three? In addition to messages in the echo area, I'm happy to work on documentation and docstrings, as well.

fommil:
Ok, let's do that

ashawley:
Ok, done

fommil:
nice, thanks again!

PR #614, opened 2017-05-08
Propertize eldoc messages.

tarao:
### Propertizing implicit information

`ensime-implicit-notes-at` returns propertized messages (in a list) but the properties were dropped because we were passing the list to `"%s"` of `format`.  `mapconcat`ing the list will preserve the properties.  We should do the same for `ensime-errors-at`, which also returns messages in a list.

### Propertizing type information

I think it is also nice to have it propertized.  Note that the local name of a symbol may be missing, which results in printing `nil: SomeType` in the original code.  We simply ignore symbols without a local name in this PR.

### DIFF
--- a/ensime-eldoc.el
+++ b/ensime-eldoc.el
@@ -17,9 +17,9 @@
   (when (ensime-connected-p)
     (let ((msg (pcase ensime-eldoc-hints
                  (`error
-                  (ensime-errors-at (point)))
+                  (mapconcat 'identity (ensime-errors-at (point)) "\n"))
                  (`implicit
-                  (ensime-implicit-notes-at (point)))
+                  (mapconcat 'identity (ensime-implicit-notes-at (point)) "\n"))
                  (`type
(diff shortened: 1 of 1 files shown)


yyadavalli:
@tarao Thanks, for the PR.

ensime-eldoc.el
```diff
@@ -17,19 +17,19 @@
...
                   (let* ((error-msg (ensime-errors-at (point)))
                          (implicit (ensime-implicit-notes-at (point)))
                          (type (ensime-eldoc-type-info)))
                     (format "%s\n%s\n%s"
-                            type
-                            (if implicit implicit "")
-                            (if error-msg error-msg "")))))))
+                            (or type "")
+                            (mapconcat 'identity implicit "\n")
+                            (mapconcat 'identity error-msg "\n")))))))
```
yyadavalli: Just a thought, is better if we get rid of the let expression.

Issue #622, opened 2017-06-01
ensime-company: can we omit these candidates, like: CollectionMapper$ListClass


yangwen0228:
### Steps to reproduce

1. Put ensime-company in the company-backends.
2. Type List

### Expected behaviour
Company candidates:
List
ListAssert
ListBuffer
...

### Actual behaviour
Company candidates:
CollectionMapper$ListClass
CollectionMapper$ListType
JSpinner$ListEditor
...
List
...

I think we can omit those candidates contain a "$" by a filter? Any suggestions is welcome! Thank you!

fommil:
(see discussion in the PR)

Issue #501, opened 2016-07-16
no ensime connection after `M-x ensime-reload`


fommil:
I was hacking on ensime-server and wanted to reload the server, but when the server came back up the emacs client was still disconnected and simple queries like `ensime-type-at-point` where saying the connection was `nil`.

PR #623, opened 2017-06-01
#622 Omit company candiates like: CollectionMapper$ListClass

yangwen0228:
#622 Omit company candidates like:
CollectionMapper$ListClass
CollectionMapper$ListType
JSpinner$ListEditor

### DIFF
--- a/ensime-completion-util.el
+++ b/ensime-completion-util.el
@@ -97,14 +97,17 @@
     (when (looking-at scala-syntax:plainid-re)
       (match-string 1))))
 
-(defun ensime-get-completions-async
-    (max-results case-sense callback)
-  (ensime-rpc-async-completions-at-point max-results case-sense
+(defun ensime-get-completions-async (max-results case-sense callback)
+  (ensime-rpc-async-completions-at-point
+   max-results case-sense
    (lexical-let ((continuation callback))
      (lambda (info)
-       (let* ((candidates (ensime--annotate-completions
-			   (plist-get info :completions))))
-	 (funcall continuation candidates))))))
+       (let* ((candidates (remove-if
+                           '(lambda (candidate)
+                              (string-match "\\$" (getf candidate :name)))
+                           (plist-get info :completions)))
+              (candidates (ensime--annotate-completions candidates)))
+         (funcall continuation candidates))))))
 
 (defun ensime-get-completions (max-results case-sense)
   (let* ((info
--- a/ensime-refactor.el
+++ b/ensime-refactor.el
@@ -44,18 +44,19 @@
   "Key bindings for the refactor confirmation popup.")
 
 (defun ensime-refactor-organize-java-imports ()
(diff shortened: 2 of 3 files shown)


fommil:
Cool! But I think we probably want to do this server side rather than hacking it on the client. Do you need help to see where to do this? Join the `ensime/ensime-server` gitter room to ask questions and/or create a ticket on the server and we can talk there.

fommil:
in fact, this would be a good complement to https://github.com/ensime/ensime-server/issues/354 (i.e. exclusion list for completions... we could allow the user to provide a custom config file with their preferred exclusion regexes once https://github.com/ensime/ensime-server/issues/1346 is ready).

fommil:
have a look on the server tracker for "completion" to see what other ideas we've had in the past.

fommil:
in fact, we do really want this so I'll create the ticket, it'd be awesome if you could hack on it!

jozi-k:
N.B. fix for #501 was also part of this pr.

fommil:
oh, was that not merged already in https://github.com/ensime/ensime-emacs/pull/620/files ?

yangwen0228:
Thank you! Of course I'd like to join the `ensime/ensime-server` gitter room. Since last time I hack on the disconnect issue, I've learned quite a lot Scala from the book 'Programming in Scala' which you recommended. And I really like Scala. It's so concise, flexible and powerful! Sometimes, it's just like some lisp code in Java style. However, I still need some time to learn the ensime-server.
BTW, I find that sometimes, when I type "myA", ensime-company completes with: `main(args: java.lang.String[])`. And sometimes other things like `finalize ...`. It's very weird. So I wrote this to exclude candidates that not begins with prefix:
```
(defun ensime-get-completions-async
    (max-results case-sense callback)
  (ensime-rpc-async-completions-at-point
   max-results case-sense
   (lexical-let ((continuation callback))
     (lambda (info)
       (let* (name
              (prefix (ensime-completion-prefix-at-point))
              (candidates (remove-if
                           '(lambda (candidate)
                              (setq name (getf candidate :name))
                              (or (string-match "\\$" name)
                                  (not (string-match prefix name))))
                           (plist-get info :completions)))
              (candidates (ensime--annotate-completions candidates)))
         (funcall continuation candidates))))))
```

fommil:
That's very weird. If you could write a test in the server that would be useful. I wonder if you are getting yasnippet completions by accident.

yangwen0228:
It's weird. Finally, I find the reason. If the code contains Chinese chars, even in the comments. This phenomenon will occur. I think this is caused by the different calculation methods of string length between English and Chinese. And the results is as follows:

fommil:
Oh I've seen that too... UTF8 in comments can mess up the compiler's range calcs. Maybe specifying the encoding would help but otherwise you may need to fix upstream. I can help with that. Raise a scala bug with a minimal repro and copy me in.

PR #625, opened 2017-06-10
remove the typecheck-all command (use source mode)

fommil:
in prep for https://github.com/ensime/ensime-server/pull/1786


ensime-editor.el
```diff
@@ -454,29 +454,6 @@ Goes to the point of the definition of the type."
   (ensime-write-buffer nil t)
   (ensime-typecheck-current-buffer))
 
-(defun ensime-reload-open-files ()
```
fommil: hmm, we still want this one. Which means we need `unload-all`

Issue #638, opened 2017-07-03
ensime-show-uses-of-symbol-at-point is not working


hzenginx:
`ensime-show-uses-of-symbol-at-point` is not working with following message:
`Symbol’s function definition is void: ensime-popup-buffer-p`

Possible reason: function `ensime-popup-buffer-p` is deleted in https://github.com/ensime/ensime-emacs/pull/631

There may be other broken functions using same function.

Reproduction:
- In any scala project, use `ensime-show-uses-of-symbol-at-point`

- [x] I have followed http://ensime.org/editors/emacs/troubleshooting/
- [x] I have followed the troubleshooting guide for my build tool
- [x] I have asked in http://gitter.im/ensime/ensime-emacs
- [ ] I have included a link to an example project
- [x] I have included exact bug reproduction instructions

PR #639, opened 2017-07-03
fix ensime-show-uses-of-symbol-at-point (#638)

hzenginx:
this is fixing #638 

this was the only usage of `ensime-popup-buffer-p` and as it is directly related about popup I've just copied function from deleted file.


fommil:
Cool, thanks.

Issue #641, opened 2017-07-10
imenu support is not working


hzenginx:
This is also because of #631 

Reproduction:
 - Open any scala project
 - try opening imenu, `popup-imenu` etc.

Error:
```
In ‘Imenu’ source: ‘helm-imenu-candidates’ 
 (void-function esnsime-imenu-index-function)
```

- [x] I have followed http://ensime.org/editors/emacs/troubleshooting/
- [x] I have followed the troubleshooting guide for my build tool
- [ ] I have asked in http://gitter.im/ensime/ensime-emacs
- [x] I have included a link to an example project
- [x] I have included exact bug reproduction instructions

PR #642, opened 2017-07-10
fix imenu support #641

hzenginx:
Solves #641 

I've just copied the old required methods to this file. looks like functions defined in `inspector.el` was also used for imenu support.


fommil:
cool, thanks

PR #648, opened 2017-07-28
Find uses of symbol at point

bomgar:
It actually does something. (Helm only)

The code is not final yet. I create this pull request so anyone else working on it can see what i've done so far.

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -985,7 +985,7 @@ copies. All other objects are used unchanged. List must not contain cycles."
 (defun ensime-rpc-uses-of-symbol-at-point ()
   (ensime-eval
    `(swank:uses-of-symbol-at-point
-     ,(buffer-file-name-with-indirect)
+     ,(ensime-src-info-with-contents-in-temp)
      ,(ensime-computed-point)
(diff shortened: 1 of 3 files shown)


yyadavalli:
@bomgar Do you think we can generalize the format function? If yes, we might be able to reuse the formatted collection with `completing-read` or `ido` for classic and `ivy-read` for ivy.

bomgar:
That should be possible. 
Also it would be awesome to render the line where the usage was found.

fommil:
you're just trying to get me to use Helm, aren't you!

fommil:
"classic" should just dump out to a grep buffer

bomgar:
I hope I can finish it this weekend. Maybe I add the classic  version as well 😄

fommil:
that'd be super awesome. BTW, please check the protocol version string

bomgar:
I also see the benefit a compile buffer provides. Maybe a fallback to the classic version with the universal-argument?

fommil:
so many of our features should be rewritten to use that. `ensime-search` should work this way too.

PR #649, opened 2017-07-29
classic show uses and type hierarchy

bomgar:
Not particularly pretty but it works.

I tried to write a feature file for the formatter but ecukes complains about a missing method.
`Symbol’s value as variable is void: imenu-auto-rescan`

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -989,6 +989,13 @@ copies. All other objects are used unchanged. List must not contain cycles."
      ,(ensime-computed-point)
      )))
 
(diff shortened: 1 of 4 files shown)


fommil:
oh, we should also think up some new keyboard shortcuts for the new hierarchy command and add it to the ensime.org docs! (both the editors tickbox page, and the emacs cheatsheet).

fommil:
how about `C-c C-v h` ?

fommil:
(I added current state of play to the docs)

fommil:
I added the binding

PR #652, opened 2017-07-31
[fix][test] fix ecukes tests

bomgar:
With this change the the ecukes test run without an error. I don't know if this is just a problem on my machine. 

What happend to the CI build?


fommil:
CI died a long time ago. I don't have the time to maintain it.

PR #653, opened 2017-07-31
[build] mark integration tests as unmaintained

bomgar:


fommil:
cool, I'll merge and enable the CI again. It might be red, but at least the fix shouldn't be too hard.

PR #654, opened 2017-08-01
[fix][feat] adapt to server changes. show preview of uses.

bomgar:
I'm really tired right now. If you want to merge it before tomorrow you should carefully review it 😉


fommil:
Niice :-)

Will merge at the same time as the server

ensime-model.el
```diff
@@ -180,6 +180,14 @@
 (defun ensime-type-fqn (type)
   (plist-get type :fqn))
 
+(defun ensime-type-source-position (type)
```
fommil: I'm generally not very keen on these things... I'd rather we used common lisp style destructuring where possible. They are just repetitive.

fommil:
Was working very well for me today!

yyadavalli:
@bomgar I'm having trouble with this. I am getting empty results (like below with classic)
```
:?:
```
any thoughts?

fommil:
what does the server response look like? (`ensime-events` may need to be enabled)

bomgar:
I just updated to the merged version and it seems to work. 
The server response would be helpful.

I'm having some results with no line or file. (e.g. java.lang.Object I guess because I don't have the sources on this machine)

fommil:
Or sometimes if something is referenced in a part of the code that has no line info, e.g. imports.

Issue #656, opened 2017-08-02
would be nice if the Hierarchy / Usages screens had a default directory


fommil:
Like grep / ag, to avoid long file paths.

// @bomgar

PR #659, opened 2017-08-07
[feat] shorten filename for uses and hierarchy

bomgar:


bomgar:
see https://github.com/ensime/ensime-emacs/issues/656

bomgar:
I'm not sure if `project-roots` is the correct method here

ensime-editor.el
```diff
@@ -897,11 +897,15 @@ falls back to the classic version."
...
 (defun ensime-format-source-position (source-position)
   "Format source position SOURCE-POSITION."
-    (let* ((file-name (ensime-pos-file source-position))
-           (maybe-line (ensime-pos-line source-position)))
-      (let ((line (if maybe-line (number-to-string (if (= 0 maybe-line) 1 maybe-line)) "?")))
-       (concat file-name
-       (propertize (concat ":" line) 'face 'font-lock-comment-face)))))
+  (let* ((file-name (ensime-pos-file source-position))
+         (maybe-line (ensime-pos-line source-position))
+         (roots (if (fboundp 'project-roots)(project-roots (project-current)) '()))
```
fommil: you could the `:root-dir` from the `.ensime`

bomgar:
`ensime-configured-project-root` works. thanks.

bomgar:
Should I change something else or is this good to go?

fommil:
oh sorry I didn't realise you'd pushed, no need to squash anymore I can do it on merge :smile: 

awesome stuff, looking forward to using it.

fommil:
@bomgar I forgot about this and upgraded and when I got results back today I was like "woah, what happened here? This is cool!" :smile:

fommil:
the hierarchy summary seems to be showing the FQN that is referenced now rather than the code snippet. Was that intentional? It is actually very useful. Or did I just not notice before because of the noise?

bomgar:
That was intentional. The server api does not return a preview for the hierarchy.

fommil:
Very nice.

PR #666, opened 2017-09-20
[feat] add ensime static search

bomgar:
Shows ensime search results in a grep buffer.

If this works good I would use the universal argument for ensime-search to trigger this.

Alternatively we could replace the classic-search with this. Users would lose interactive search though.


fommil:
You're a star!

bomgar:
Universal argument to ensime-search now triggers static search

fommil:
I'd be happy with this as default, deleting the old custom mode. And autopopulate the search box with thing at point. Great M-. Failure backup

PR #667, opened 2017-09-20
static search is awesomesauce, make it default

fommil:


fommil:
working on a couple of bugs I added...

bomgar:
I would still like to use classic with the universal argument to ensime-search.

Impressive how much code the custom mode had.

fommil:
Oops... should we bring it back then? I'm definitely in favour of scrapping the custom code and using ivy or helm (or another?)

PR #668, opened 2017-09-21
[feat] use universal argument to force classic search

bomgar:


bomgar:
I guess I should have called it 'new classic search'

PR #670, opened 2017-10-04
Fixing ensime-helm-select-source-position to display previews

tbje:
Fixing `ensime-helm-select-source-position` - (`C-c C-v r` with `ensime-search-interface` = `'helm` so that previews are displayed.

helm-with

instead of 

helm-without

### DIFF
--- a/ensime-helm.el
+++ b/ensime-helm.el
@@ -9,9 +9,19 @@
 
 (require 'helm "helm.el" t)
 
+(defun string-or-empty (str)
+  "Returns the string or empty string if nil"
+  (if str str ""))
+
 (defun ensime-helm-select-source-position (positions name)
   "Select one source position element using helm"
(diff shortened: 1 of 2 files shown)


ensime-helm.el
```diff
@@ -9,9 +9,19 @@
 
 (require 'helm "helm.el" t)
 
+(defun string-or-empty (str)
```
bomgar: Functions should always have an ensime prefix. Elisp lacks namespaces. Maybe instead of this function `(or str "" )` is good enough.

tbje: Thanks! Sounds like an excellent improvement :)

bomgar (approved):

fommil:
@bomgar please feel free to merge when you're happy with something.

fommil:
(the world explodes, @bomgar instantly regrets everything)

PR #673, opened 2017-10-11
[WIP] Add ensime-api.el: a mirror of the server-side API

iravid:
@fommil @bomgar @yyadavalli - would be good to get your feedback before continuing further

### DIFF
--- a/Cask
+++ b/Cask
@@ -32,6 +32,7 @@
  (depends-on "auto-complete")
  (depends-on "helm")
  (depends-on "ivy")
+ (depends-on "eieio")
  ;;(depends-on "expand-region")
  (depends-on "f")
  (depends-on "ert-runner")
--- a/ensime-api.el
+++ b/ensime-api.el
@@ -1,0 +1,315 @@
+(require 'eieio)
+
+(ensime-marshal-defclass ensime-api-offset-range ()
+  ((from :initarg :from
+         :type number)
+   (to :initarg :to
+       :type number)))
+
+(ensime-marshal-defclass ensime-api-implicit-info-req ()
+  ((file :initarg :file
+         :type string)
+   (range :initarg :range
+          :type ensime-api-offset-range)))
+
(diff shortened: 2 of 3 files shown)


fommil:
This is so cool, I've wanted to do this for so long. Maybe getting into the disciple of having test data for unit tests would be good early on?

I recommend deleting all the other code from 3.0, we can bring it back as needed. Might help focus the mind on core functionality. If be really keen on seeing a small core with a contrib bizzarre.

iravid:
Thanks Sam. Tests will follow shortly; I plan on mirroring SwankyFormatsSpec.

________________________________

bomgar:
Looks awesome. The only sad part is that all contributed features need to be rewritten.

fommil:
I kinda feel like the codebase was becoming an untamed jungle so I'm pretty pleased that it means we get to go back to basics and rethink "what do we actually need / use?" :smile:

iravid:
I agree that it is unfortunate to rewrite everything, but the codebase has become quite a zoo. There's also no rush as 2.0 works well and we can take our time in refining the 3.0 branch.

iravid:
Anyway, I am working on adding coproduct discriminator support to `marshal.el`; will continue this branch once I finish that.

iravid:
Sometimes the path of least resistance is just a homegrown implementation. I gave up on marshal.el after struggling to add the discriminator support.

I added `ensime-marshal.el` that does just what we need - marshalling/unmarshalling to/from SWANKY.

fommil:
I wonder if macros could be used to make the runtime super fast :smile:

iravid:
We definitely could use macros to generate all the serde code. Is it worth our time? Not sure ;) I still have all of `outgoing.scala` to translate.

iravid:
Hey, maybe I should use `scalameta` to do the translation ;)

fommil:
Maybe we should have something generate the Scala from the lisp :-P

PR #674, opened 2017-10-17
[wip] test websocket connection

bomgar:
@iravid @fommil I'm currently playing around with the websocket library. Currently the server responds with 

```
ws frame: "(
  :payload (
    :ensime-api-ensime-server-error (
      :description \"org.ensime.sexp.DeserializationException: read should never be called for CNil, ListMap()\"
    )
  )
)"
```

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -45,6 +45,7 @@
   (require 'ensime-macros))
 (require 'ensime-vars)
 (require 'dash)
+(require 'ensime-websocket)
 
 (defvar ensime-net-processes nil
   "List of processes (sockets) connected to Lisps.")
@@ -308,7 +309,7 @@ This doesn't mean it will connect right after Ensime is loaded."
     process))
 
 (defun ensime-connect (host port &optional disconnect)
-  "Connect to a running Swank server. Return the connection."
+  "Connect to a running SWANKY server. Return the connection."
   (interactive (list
         (read-from-minibuffer "Host: " "127.0.0.1")
         (read-from-minibuffer "Port: " nil nil t)
@@ -316,8 +317,8 @@ This doesn't mean it will connect right after Ensime is loaded."
 	     (y-or-n-p "Close old connections first? "))))
   (when disconnect
     (ensime-disconnect-all))
-  (message "Connecting to Swank on port %S.." port)
-  (let* ((process (ensime-net-connect host port))
+  (message "Connecting to SWANKY on port %S.." port)
+  (let* ((process (ensime-websocket-connect host port))
          (ensime-dispatching-connection process))
     (ensime-setup-connection process)))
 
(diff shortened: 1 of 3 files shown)


iravid:
Very happy to see that you’re picking this up :-)

bomgar:
Now I actually get a valid response. I'm going to build the glue code to the server starter next.

Is my understanding correct that the with the websocket api we need to store callbacks for the `call-id` and call them async when the response arrives?

fommil:
That should be the same as the old protocol. Sounds about right.

iravid:
Yeah, I think this is the part that dispatches events to continuations: https://github.com/ensime/ensime-emacs/blob/2.0/ensime-client.el#L640

fommil:
this is why we rock vs LSP :smiley: (they just block)

iravid:
Really? That’s terrible

fommil:
the protocol might have support, but the emacs impls don't use it.

bomgar:
Starting and connecting to the server is a lot more code than I had anticipated.
I think I now have a basic understanding of what it does. 
The code now connects to the new server but doesn't realize it is connected so it establishes a connection every 5 seconds.

bomgar:
The current implementation assumes every network connection has a buffer associated with it. It is really confusing that a lot of methods distinguish between a buffer and a connection. 

I think I give up on recycling the current code and rewrite something more suitable for websockets.

The network buffer is used by the ensime-macros to store variables into them. This is something that doesn't really work with a websocket connection. I wonder if we could use the ensime server buffer to store the websocket connection.

yyadavalli:
@bomgar I think that's a good idea. I'm very much leaning towards @fommil's suggestion of starting fresh with 3.0

fommil:
with unit tests and coverage reports from the beginning :smiley:

iravid:
Agreed. Delete it all and rewrite. :-)

yyadavalli:
Cool. I'll do the purge and @bomgar you can rebase the PR whenever you are ready.

bomgar:
No promises. I try to find the time but if someone is faster than me I'm happy 😄

chessman:
@fommil @yyadavalli 
Hi. I'm trying to continue this work and need some elisp advises. What is the best way to destructure plists? I just use it like O(n)-map now. Does the server guarantees the order of properties?

fommil:
@chessman I'm not involved with ensime anymore (I'm not even a user), so I'm afraid I can't help out... but I do know the answer to your `plist` question :smile: Have a look at recent additions to this wonderful library https://github.com/magnars/dash.el/issues/111 (probably only in melpa unstable)

chessman:
Oh, nice. `-when-let*` can be used like for-comprehension with `Option`. Thank you.

PR #676, opened 2017-11-08
The Purge

yyadavalli:


fommil (approved):

tarsius:
Please consider renaming the **branch** `2.0` to `master` and `3.0` to `next`.

yyadavalli:
@tarsius do you have a specific reason for that. We have it as `3.0` and `2.0` to align with [ensime-server](https://github.com/ensime/ensime-server)

tarsius:
As the maintainer of the Emacsmirror I run into many issues where someone decides to break with the conventions, such as having a branch named `master`, or following elisp header conventions, or not bundling third-party libraries, or not using homegrown licenses. Usually the maintainers have, or believe to have, some good reason to depart from the conventions, but often it just means that others will have to deal with those idiosyncrasies.

I have already done this in this case. So I don't don't mind if you keep things as-is - I just though I would point out that this is not optimal. It would actually be more work for me if you did revert to follow the conventions (but not very much).

yyadavalli:
Sure, i'll wait for the other maintainers to comment and rename based on their opinion. I'll ping you if we do so. Let me know if I can do something to make is easy for you to account for the changes.

/cc @fommil @bomgar @iravid

tarsius:
It's not that what you have done doesn't make sense. So stick to it if you prefer. Really.

> Let me know if I can do something to make is easy for you to account for the changes.

No additional action needed - I will see how this is resolved.

iravid:
@yyadavalli @tarsius My only opinion here is playing nice with the ecosystem. I'd rather see more magit features rather than have @tarsius spend time on workarounds ;) 

`master`/`2.0` and `next`/`3.0` are all the same to me :-)

fommil:
I think that will be automatically compatible with melpa, if anybody wants to make the necessary changes. Just make sure `master` is the "default" branch.

yyadavalli:
Since changing branch would only add more work for @tarsius at this point, I'm leaving them as is

PR #687, opened 2018-03-08
Update README.md,add Contribution link

doofin:


fommil:
Sorry, no. I don't care about impatient people. We have extensive docs at http://ensime.org

Issue #690, opened 2018-03-15
yasnippet expansion doesn't work 


nverno:
Leave this list and include an `x` in the relevant boxes:

- [x] I have followed http://ensime.org/editors/emacs/troubleshooting/
- [x] I have followed the troubleshooting guide for my build tool
- [ ] I have asked in http://gitter.im/ensime/ensime-emacs
- [ ] I have included a link to an example project
- [ ] I have included exact bug reproduction instructions

nverno:
The fallback behaviour for tab-completion doesn't work as expected in ensime. 
After loading ensime, TAB no longer completes snippets in any mode. There is 
something wrong with the way it is being set in `company-ensime-enable`

PR #691, opened 2018-03-15
Company and yasnippet

nverno:
Currently ensime-company-enable rebinds yas-minor-mode keys, which stops snippets
from expanding.  I'm not sure of the deeper cause, but this seems to fix things as far as 
I can tell. #690

### DIFF
--- a/ensime-company.el
+++ b/ensime-company.el
@@ -11,7 +11,7 @@
 ;;; Code:
 
 (eval-when-compile
-  (require 'cl)
+  (require 'cl-lib)
   (require 'ensime-macros))
 
 (require 'ensime-client)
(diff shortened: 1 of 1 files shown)


fommil:
can you please recreate this without the cl-lib and formatting changes? I want to minimise the diff. I'd be happy to accept the cleanups in a follow up PR.

fommil:
I'm not sure about the yasnippet changes. I'd be concerned that it breaks what we currently have. Certainly the yasnippet TAB story is not great, have you followed the advice in the Hacks page?

nverno:
yea sure will do. I would be curious why those changes were added in the first place as they seem to break yasnippet for me in all modes. The binding to `local-keymap` is particularly suspect

fommil:
I'm declaring all bugs bankrupt, but if you pick up this PR again please reopen it! :smile:

PR #692, opened 2018-03-17
Remove call to non-existent 'ensime-sbt-do-scalariform-only'

mmajcher:
'ensime-sbt-do-scalariform-only' has been removed by:
3d3ab18436ad6089496b3bce1d49c64a86965431

### DIFF
--- a/ensime-editor.el
+++ b/ensime-editor.el
@@ -700,15 +700,6 @@ If user selects an import, add it to the import list."
               (ensime-typecheck-current-buffer)))))
       (message "No import suggestions were returned for %S" name))))
 
-;; Source Formatting - transition cue to sbt task
-(defun ensime-format-source ()
-  "DEPRECATED - use ensime-sbt-do-scalariform-only directly.
-Functionality was moved from ensime-server to the build tool plugins.
-Use build tools tasks appropriately"
-  (interactive)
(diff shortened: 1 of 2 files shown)


ensime-editor.el
```diff
@@ -706,8 +706,7 @@ If user selects an import, add it to the import list."
 Functionality was moved from ensime-server to the build tool plugins.
 Use build tools tasks appropriately"
   (interactive)
-  (message "Function use is deprecated, please transition to ensime-sbt-do-scalariform-only.")
-  (ensime-sbt-do-scalariform-only))
+  (message "Function use is deprecated, please transition to ensime-sbt-do-scalariform-only."))
```
iravid: Would it make sense to get rid of `ensime-format-source` entirely?

mmajcher:
Removed all of 'scalariform' references

iravid:
Cool. Any objections? cc @fommil @yyadavalli

fommil:
cool. Generally in 3.0 I'd like us to move towards a small kernel that understands the `.ensime` file and server protocol, and have a separate "contrib" module for anything like this.

I've also asked the sbt-scalariform maintainers to take ownership of the `scalariformOnly` task that we used to have, but they don't seem to want it. I don't care because I've moved onto scalafmt for everything.

Issue #662, opened 2017-09-06
C-c C-z C-r eval is going into paste mode and getting stuck


fommil:

cmears:
It seems like Emacs sends the EOF too fast, before the interpreter is in "wait for EOF" mode. I found that adding `(sleep-for 0.2)` before the `(comint-send-eof)` in `ensime-inf-eval-region` is a viable work-around.

Issue #588, opened 2017-03-29
ensime-inf-eval-region requires ctrl-D


lambda-knight:
Hi,
In my Emacs environment, ensime-inf-eval-region requires ctrl-D in a REPL screen 
to start eval. Is it correct behavior ?  I'd like to eval the region in Scala code automatically 
when I issue ensime-inf-eval-region. The following is a screen image of emacs. REPL is 
waiting for ctrl-D.

![image](https://cloud.githubusercontent.com/assets/25827631/24479290/7462db68-1519-11e7-9e90-5b829ba3881a.png)

The version of ensime is 20170318.1059.

lambda-knight:
I tested again. It works sometime and it doesn't work sometime. I suspect my emacs setting causes the problem. I close this issue.

Thanks

fommil:
If you can reliably reproduce, please open up a ticket. I sometimes see it too but I don't know why.

lambda-knight:
Sure, I see it again but I don't know when that occurs.

fommil:
I'm guessing it has to do with newlines.

mashu:
I have the same problem, C-c C-v r asks me to press C-d in the REPL, then the code gets executed.
This is not usable, and since more people report this, this should be investigated. I doubt we all use the same "environment".

**It is also reported in separate issues "duplicates" of this!!**
For example https://github.com/ensime/ensime-emacs/issues/662

fommil:
@mashu glad to hear that you plan on investigating this.

mashu:
You are welcome, **steps to reproduce**

1. Create empty ~/.emacs.d
2. Use most basic init.el from http://ensime.org/editors/emacs/learning/
3. Follow tutorial http://ensime.org/editors/emacs/install/ adding three packages from maple-stable
4. C-x C-f to open scala file with just println("Hello world") example
5. M-x ensime
6. C-c C-v z to open REPL
7. Select region with emacs and press C-c C-v b which should evaluate region

Instead you will see in REPL this
`
Welcome to Scala 2.12.4 (OpenJDK 64-Bit Server VM, Java 1.8.0_151).
Type in expressions for evaluation. Or try :help.
scala> // Entering paste mode (ctrl-D to finish)
`

and you need to switch buffors to REPL and press C-D to get result. What is expected is to evaluate region immediately without the need of switching buffers and pressing additional keys.
`
[...]

PR #695, opened 2018-03-18
Fix 'ensime-inf-eval-region'

mmajcher:
1. Replaced 'next-line' with 'forward-line'
Next-line is meant for interactive use only. Often it would change the
position of point in current buffer (with Scala source), while it
should only affect Repl buffer.

2. Short sleep before comint-send-eof
Sometimes Emacs sends EOF to interpreter too fast and intepreter
doesn't exit paste mode (as it's supposed to do).

3. Short sleep before postoutput-filter
Sometimes Emacs looks for eval result before all of the result from
interpreter has been printed to Repl buffer. This results in sending
only part of eval result to overlay.

https://github.com/ensime/ensime-emacs/issues/662
https://github.com/ensime/ensime-emacs/issues/588

### DIFF
--- a/ensime-inf.el
+++ b/ensime-inf.el
@@ -236,6 +236,7 @@ Used for determining the default in the next one.")
       (comint-send-string nil ":paste\n")
       (comint-send-string nil reg)
       (comint-send-string nil "\n")
+      (sleep-for 0.2)
       (comint-send-eof))))
 
 (defun ensime-inf-eval-result ()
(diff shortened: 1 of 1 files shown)


iravid (approved):
Thanks @mmajcher. Assuming you've tested this locally? Let me know if so and I'll merge.

mmajcher:
`cask exec ert-runner` and `cask exec ecukes`: ok

`run_emacs_test.sh`: well...

1) had to comment `// ensimeServerVersion in ThisBuild := "2.0.0-M1"` in `test/example`
-> this results in `(error "ENSIME protocol 1.9.1 is too old, update the build tool plugin / server")`
-> once commented, it uses `2.3.0` from my global config

2) `test/run_emacs_test.sh` results in: 
```
Starting test: ’Cache index.’
ENSIME server starting...
Connecting to Swank on port 45133..
ENSIME protocol 1.9.6
ENSIME ready. Find closure!
```
... and it doesn't finish. It occupies CPU for a while. 
-> it's the same with current `2.0` branch.
-> if test `Cache index.` is commented out, the same happens with `Test edit definition.`

My emacs is 25.2.2

iravid:
Sorry, by "testing" I meant verifying that `ensime-inf-eval-region` still works properly :-)

mmajcher:
Yes, works properly :)

iravid:
Thanks!

PR #696, opened 2018-05-01
Try cooperating with ensimeServerVersion in ThisBuild := "3.0.0-SNAPSHOT"

dickmao:
Ensure `$HOME/.sbt/1.0/global.sbt` is clean of the following.  Then add to `ensime.sbt`
```
ensimeServerVersion in ThisBuild := "3.0.0-SNAPSHOT"
ensimeProjectServerVersion in ThisBuild := "3.0.0-SNAPSHOT"
```
Then `sbt ensimeConfig` and `M-x ensime` on `foo.scala`.
Output of vovapolu language server in `.ensime_cache/pc.stdout.log`

ensime/ensime-server#1935

### DIFF
--- a/ensime-client.el
+++ b/ensime-client.el
@@ -149,10 +149,9 @@ overrides `ensime-buffer-connection'.")
         conn)))
 
 (defun ensime-proc-if-alive (proc)
-  "Returns proc if proc's buffer is alive and proc has not exited,
- otherwise nil."
+  "Returns proc if proc's buffer is alive (or doesn't have an associated buffer) and proc has not exited, otherwise nil."
   (when (and proc
-	     (buffer-live-p (process-buffer proc))
+	     (or (null (process-buffer proc)) (buffer-live-p (process-buffer proc)))
 	     (let ((status (process-status proc)))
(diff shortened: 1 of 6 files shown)


fommil:
Thanks for looking at this!

I think it would make more sense to team up with https://github.com/rossabaker/lsp-scala/issues/1 for this though, do you agree?

In ensime 3.0 we plan on writing a very very minimal swanky based protocol, plus more non-server commands that use the `.ensime` / build definition. But if the scalabus comes along in time maybe we should just focus on making lsp-mode better.

dickmao:
Between the single elisp file in rossabaker/lsp-scala and the completely purged 3.0 branch in ensime/ensime-emacs, "very very minimal" seems apropos.  Just to quickly get vovapolu's server stood up for emacs, I felt reusing the existing data structures and config reader in the 2.0 branch would be most expedient but I also see why that conflates things.  Please let me know if I should close the PR.  Thanks.

fommil (changes requested):
I don't think I want to merge this, but it would be good to leave it open so that anybody who wants to join in hacking something together knows where the fun is happening.

PR #702, opened 2018-11-10
Fixes infinite loop in refactoring

what-the-functor:
Fixes bug that the refactoring functions prompt endlessly:
> Hunk has already been applied; Undo it? (y or n)

Removes iteration of diff-header search results, in favour of functionality of `diff-apply-hunk`; to apply each hunk in sequence.

Tested in Emacs 26.1

### DIFF
--- a/ensime-refactor.el
+++ b/ensime-refactor.el
@@ -229,10 +229,9 @@
   "Apply or undo all hunks in the diff contents of the current buffer."
   (interactive)
   (make-local-variable 'diff-advance-after-apply-hunk)
-  (setq diff-advance-after-apply-hunk nil)
+  (setq diff-advance-after-apply-hunk t)
   (goto-char (point-min))
-  (while (re-search-forward diff-hunk-header-re nil t)
-    (diff-apply-hunk)))
+  (diff-apply-hunk))
 
 (defun ensime-refactor-diff-save-source-files ()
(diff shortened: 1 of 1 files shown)


ensime-refactor.el
```diff
@@ -229,10 +229,9 @@
   "Apply or undo all hunks in the diff contents of the current buffer."
   (interactive)
   (make-local-variable 'diff-advance-after-apply-hunk)
-  (setq diff-advance-after-apply-hunk nil)
+  (setq diff-advance-after-apply-hunk t)
   (goto-char (point-min))
-  (while (re-search-forward diff-hunk-header-re nil t)
```
cruatta, line 234: I think the while loop is still needed. Refactoring multiple instances of something required it when I was testing.

beyondpie:
I keep the while loop, and it works OK for me now. But not test the case without while loop. Emacs version 26.1 .

beyondpie:
@cruatta 
can we merge the PR with your modification?
Or @tonylotts 
would you mind to change your code according to the suggestion? If your time is limited, can I copy your solution and send the PR?

jinschoi:
Just changing the "(setq diff-advance-after-apply-hunk t)" line works. Thanks for tracking this down, it's been bugging me forever.

macalinao:
What's the status on this?

